PR #2, opened 2020-09-29
GRIDEDIT-165: bug fix, reactivate old unit test

lucacarniato:
LandBoundaries is not the best example of clean implementation. An initialized variable causes a wrong allocation and inconsistent unit test run times. 

LandBoundaries needs to be refactored in a separate issue

### DIFF
--- a/src/LandBoundaries.cpp
+++ b/src/LandBoundaries.cpp
@@ -43,20 +43,16 @@ namespace meshkernel
         m_mesh(mesh), 
         m_polygons(polygons)
     {
-        m_numNode = 0;
-        m_numAllocatedNodes = 0;
+        m_numNode = int(landBoundary.size());
+
         if (!landBoundary.empty())
         {
             ResizeVectorIfNeededWithMinimumSize(int(landBoundary.size()), m_nodes, m_allocationSize, { doubleMissingValue,doubleMissingValue });
-            m_numAllocatedNodes = int(m_nodes.size());
-
-            m_numNode = 0;
             for (int i = 0; i < landBoundary.size(); i++)
             {
-                m_nodes[m_numNode] = landBoundary[i];
-                m_numNode++;
-            }
+                m_nodes[i] = landBoundary[i];
 
+            }
             m_polygonNodesCache.resize(maximumNumberOfNodesPerFace);
         }
 
--- a/src/LandBoundaries.cpp
+++ b/src/LandBoundaries.cpp
@@ -43,31 +43,24 @@ namespace meshkernel
         m_mesh(mesh), 
         m_polygons(polygons)
     {
-        m_numNode = int(landBoundary.size());
-
         if (!landBoundary.empty())
         {
-            ResizeVectorIfNeededWithMinimumSize(int(landBoundary.size()), m_nodes, m_allocationSize, { doubleMissingValue,doubleMissingValue });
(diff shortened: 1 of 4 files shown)


lucacarniato:
[https://issuetracker.deltares.nl/browse/GRIDEDIT-165](url)

src/LandBoundaries.cpp
```diff
@@ -43,18 +43,21 @@ namespace meshkernel
...
-        m_polygonNodesCache.resize(maximumNumberOfNodesPerFace);
+        m_numAllocatedNodes = 0;
+        if (!landBoundary.empty())
+        {
+            ResizeVectorIfNeededWithMinimumSize(int(landBoundary.size()), m_nodes, m_allocationSize, { doubleMissingValue,doubleMissingValue });
+            m_numAllocatedNodes = int(m_nodes.size());
 
-        ResizeVectorIfNeededWithMinimumSize(m_numNode + int(landBoundary.size()), m_nodes, m_allocationSize, { doubleMissingValue,doubleMissingValue });
-        m_numAllocatedNodes = int(m_nodes.size());
+            m_numNode = 0;
```
Hofer-Julian, line 53: Why do you set `m_numNode` to zero twice?

Hofer-Julian (approved):
One comment needs to be clarified, apart from that the changes seem fine.

Unrelated to that, we might want to get rid of the manual resizing of the vectors at some point.

src/LandBoundaries.cpp
lucacarniato: See my second commit (a simplification). in my first commit m_numNode needs to be assigned to a nongarbage value in case the landboundary line is empty

src/LandBoundaries.cpp
```diff
@@ -43,18 +43,17 @@ namespace meshkernel
...
-        ResizeVectorIfNeededWithMinimumSize(m_numNode + int(landBoundary.size()), m_nodes, m_allocationSize, { doubleMissingValue,doubleMissingValue });
-        m_numAllocatedNodes = int(m_nodes.size());
-
-        for (int i = 0; i < landBoundary.size(); i++)
+        if (!landBoundary.empty())
         {
-            m_nodes[m_numNode] = landBoundary[i];
-            m_numNode++;
+            ResizeVectorIfNeededWithMinimumSize(int(landBoundary.size()), m_nodes, m_allocationSize, { doubleMissingValue,doubleMissingValue });
+            for (int i = 0; i < landBoundary.size(); i++)
```
Hofer-Julian, line 51: Here would be an iterator nice.
I think that would also fix one code smell.

src/LandBoundaries.cpp
```diff
@@ -148,29 +139,26 @@ namespace meshkernel
...
+                m_segmentIndices.push_back({ startInd, endInd });
             }
             begin = found + 1;
         }
 
         // Generate two segments for closed land boundaries
-        int numSegmentIndexsesBeforeSplitting = m_numSegments;
-        for (int i = 0; i < numSegmentIndexsesBeforeSplitting; i++)
+        const auto numSegmentIndexsesBeforeSplitting = m_segmentIndices.size();
+        for (size_t i = 0; i < numSegmentIndexsesBeforeSplitting; i++)
```
Hofer-Julian, line 153:
```suggestion
        for (auto i = 0; i < numSegmentIndexsesBeforeSplitting; i++)

```

PR #4, opened 2020-09-30
GRIDEDIT-166: Fix `mkernel_deallocate_state`

Hofer-Julian:
I wonder if `meshInstances` should be a linked list instead of a vector.
`erase` is quite expensive if not done at the end of the vector.

Please see https://issuetracker.deltares.nl/browse/GRIDEDIT-166


lucacarniato:
ok

PR #17, opened 2020-10-19
GRIDEDIT-209: Move from error codes to exceptions

Hofer-Julian:
See https://issuetracker.deltares.nl/browse/GRIDEDIT-209


lucacarniato (approved):
already reviewed

PR #28, opened 2020-11-19
GRIDEDIT-240: Fix problems introduced with refactoring

Hofer-Julian:
+ use VERSION_SUFFIX instead of BUILD_NUMBER. VERSION_SUFFIX is retrieved from the environment
+ VERSION_SUFFIX is added only for development NuGet package


lucacarniato (approved):

lucacarniato:
+ use VERSION_SUFFIX instead of BUILD_NUMBER. VERSION_SUFFIX is retrieved from the environment
+VERSION_SUFFIX is added only for development NuGet package

PR #42, opened 2020-12-07
GRIDEDIT-259: Fix code smells at API level

Hofer-Julian:
Also update version number since this is a breaking change


lucacarniato (approved):
Al good, unused parameters removed

PR #65, opened 2021-02-02
GRIDEDIT-291: Make sure that we can compile MeshKernel on CentOS 7

lucacarniato:


.gitignore
```diff
@@ -21,3 +21,4 @@ xbuild
 
 # User specific
 post_build.bat
+/tests/utils/include/TestUtils/Definitions.hpp
```
Hofer-Julian, line 24: What is that one good for?

CMakeLists.txt
```diff
@@ -6,8 +6,8 @@ project(
   MeshKernel
   VERSION 0.10.0
   DESCRIPTION "Library for creating and editing meshes."
-  LANGUAGES CXX)
-
+  LANGUAGES CXX C)
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
Hofer-Julian, line 10: Why is this necessary?

CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
     set(CMAKE_CXX_FLAGS_RELEASE
         "-g -O2"
         CACHE STRING "List of C++ compiler flags for a Release build")
```
Hofer-Julian, line 19: Why don't you specify c++17 in the release build as well?

CMakeLists.txt
```diff
@@ -44,9 +44,23 @@ add_subdirectory(extern)
...
+endif()  
+if(WIN32)
+  # In windows use static libraries
+  set(Boost_USE_STATIC_LIBS ON)
+  find_package(Boost REQUIRED)
+endif()  
 include_directories(${Boost_INCLUDE_DIR})
 
+#NetCDF
+find_path(NETCDF_BIN_PATH NAMES "${CMAKE_SHARED_LIBRARY_PREFIX}netcdf${CMAKE_SHARED_LIBRARY_SUFFIX}"
```
Hofer-Julian, line 59: We do this only on Windows, right?
So make add an if(Win32) to it

src/MeshKernelApi/MeshKernel.cpp
```diff
@@ -123,8 +124,7 @@ namespace meshkernelapi
 
         for (const auto& index : indices)
         {
-            const auto size = index[1] - index[0] + 1;
-            if (size > 0)
+            if (const auto size = index[1] - index[0] + 1; size > 0)
```
Hofer-Julian, line 127: I still don't like this syntax...

Hofer-Julian (commented):
Just a few small comments

CMakeLists.txt
lucacarniato: bcs triangle is in C, and g++ gnu compiler seems not able to compile .c files. we need to explicitly state is also a c project

CMakeLists.txt
lucacarniato: good point, i will add it

src/MeshKernelApi/MeshKernel.cpp
lucacarniato: I see your point if I can't inspect the value of size when debugging. if this is the case we can revert this

CMakeLists.txt
Hofer-Julian: I meant the "CMAKE_POSITION_INDEPENDENT_CODE ON" part.
Could you please elaborate on that?

PR #75, opened 2021-03-03
GRIDEDIT-333: add curvilinear grid derefinement

lucacarniato:
…rom CurvilinearGrid class

### DIFF
--- a/src/MeshKernel/CurvilinearGridDeRefinement.cpp
+++ b/src/MeshKernel/CurvilinearGridDeRefinement.cpp
@@ -1,0 +1,89 @@
+//---- GPL ---------------------------------------------------------------------
+//
+// Copyright (C)  Stichting Deltares, 2011-2021.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation version 3.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+// contact: delft3d.support@deltares.nl
+// Stichting Deltares
+// P.O. Box 177
+// 2600 MH Delft, The Netherlands
+//
+// All indications and logos of, and references to, "Delft3D" and "Deltares"
+// are registered trademarks of Stichting Deltares, and remain the property of
+// Stichting Deltares. All rights reserved.
+//
+//------------------------------------------------------------------------------
+
+#pragma once
+
+#include <utility>
+
+#include <MeshKernel/CurvilinearGrid.hpp>
+#include <MeshKernel/CurvilinearGridDeRefinement.hpp>
+#include <MeshKernel/Entities.hpp>
+#include <MeshKernel/Splines.hpp>
+
+meshkernel::CurvilinearGridDeRefinement::CurvilinearGridDeRefinement(std::shared_ptr<CurvilinearGrid> grid, const Point& firstPoint, const Point& secondPoint)
+    : m_grid(std::move(grid)),
+      m_firstPoint(firstPoint),
+      m_secondPoint(secondPoint)
+{
+}
+
+meshkernel::CurvilinearGrid meshkernel::CurvilinearGridDeRefinement::Compute()
+{
+    // Get the m and n indices from the point coordinates
+    auto [mFirstNode, nFirstNode] = m_grid->GetNodeIndices(m_firstPoint);
+    auto [mSecondNode, nSecondNode] = m_grid->GetNodeIndices(m_secondPoint);
+
+    /// The points must lie on the same gridline
+    if (mSecondNode - mFirstNode != 0 && nSecondNode - nFirstNode != 0)
+    {
+        throw std::invalid_argument("CurvilinearGridDeRefinement::Compute: The selected curvilinear grid nodes are not on the same grid-line");
+    }
+
+    /// estimate the dimension of the refined grid
+    const auto numMToDeRefine = mSecondNode - mFirstNode == 0 ? 1 : mSecondNode - mFirstNode;
+    const auto numNToDeRefine = nSecondNode - nFirstNode == 0 ? 1 : nSecondNode - nFirstNode;
+
+    // the de-refined grid
+    std::vector<std::vector<Point>> deRefinedGrid;
+    deRefinedGrid.reserve(m_grid->m_numM);
+
+    size_t localMDeRefinement;
+    for (auto m = 0, mIndexOriginalGrid = 0; m < m_grid->m_numM - 1, mIndexOriginalGrid < m_grid->m_numM; ++m, mIndexOriginalGrid += localMDeRefinement)
+    {
+        localMDeRefinement = 1;
+        if (mIndexOriginalGrid >= mFirstNode && mIndexOriginalGrid < mSecondNode)
+        {
+            localMDeRefinement = numMToDeRefine;
+        }
+
+        deRefinedGrid.emplace_back(std::vector<Point>());
+        deRefinedGrid.back().reserve(m_grid->m_numN);
+        size_t localNDeRefinement;
+        for (auto n = 0, nIndexOriginalGrid = 0; n < m_grid->m_numN - 1, nIndexOriginalGrid < m_grid->m_numN; ++n, nIndexOriginalGrid += localNDeRefinement)
+        {
+
+            localNDeRefinement = 1;
+            if (nIndexOriginalGrid >= nFirstNode && nIndexOriginalGrid < nSecondNode)
+            {
+                localNDeRefinement = numNToDeRefine;
+            }
+            deRefinedGrid.back().emplace_back(m_grid->m_gridNodes[mIndexOriginalGrid][nIndexOriginalGrid]);
+        }
+    }
+    // substitute original grid with the derefined one
+    return CurvilinearGrid(deRefinedGrid, m_grid->m_projection);
+}
--- a/src/MeshKernel/CurvilinearGridDeRefinement.cpp
+++ b/src/MeshKernel/CurvilinearGridDeRefinement.cpp
@@ -62,7 +62,7 @@ meshkernel::CurvilinearGrid meshkernel::CurvilinearGridDeRefinement::Compute()
     deRefinedGrid.reserve(m_grid->m_numM);
 
     size_t localMDeRefinement;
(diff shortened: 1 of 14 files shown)


src/MeshKernel/CurvilinearGridDeRefinement.cpp
```diff
@@ -0,0 +1,89 @@
...
+    auto [mSecondNode, nSecondNode] = m_grid->GetNodeIndices(m_secondPoint);
+
+    /// The points must lie on the same gridline
+    if (mSecondNode - mFirstNode != 0 && nSecondNode - nFirstNode != 0)
+    {
+        throw std::invalid_argument("CurvilinearGridDeRefinement::Compute: The selected curvilinear grid nodes are not on the same grid-line");
+    }
+
+    /// estimate the dimension of the refined grid
+    const auto numMToDeRefine = mSecondNode - mFirstNode == 0 ? 1 : mSecondNode - mFirstNode;
```
Hofer-Julian, line 57: The use of min, would probably be more readable

src/MeshKernel/CurvilinearGridDeRefinement.cpp
```diff
@@ -0,0 +1,89 @@
...
+    const auto numNToDeRefine = nSecondNode - nFirstNode == 0 ? 1 : nSecondNode - nFirstNode;
+
+    // the de-refined grid
+    std::vector<std::vector<Point>> deRefinedGrid;
+    deRefinedGrid.reserve(m_grid->m_numM);
+
+    size_t localMDeRefinement;
+    for (auto m = 0, mIndexOriginalGrid = 0; mIndexOriginalGrid < m_grid->m_numM; ++m, mIndexOriginalGrid += localMDeRefinement)
```
Hofer-Julian, line 65: This for loop should be cleaned up a bit.
Sonarcloud gives helpful suggestions here

src/MeshKernel/CurvilinearGridDeRefinement.cpp
```diff
@@ -0,0 +1,89 @@
...
+    {
+        localMDeRefinement = 1;
```
Hofer-Julian, line 67: You probably want to init the var to 1

src/MeshKernel/CurvilinearGridDeRefinement.cpp
```diff
@@ -0,0 +1,89 @@
...
+        if (mIndexOriginalGrid >= mFirstNode && mIndexOriginalGrid < mSecondNode)
+        {
+            localMDeRefinement = numMToDeRefine;
+        }
+
+        deRefinedGrid.emplace_back(std::vector<Point>());
+        deRefinedGrid.back().reserve(m_grid->m_numN);
+        size_t localNDeRefinement;
+        for (auto n = 0, nIndexOriginalGrid = 0; nIndexOriginalGrid < m_grid->m_numN; ++n, nIndexOriginalGrid += localNDeRefinement)
```
Hofer-Julian, line 76: Please clean up this for loop as well

src/MeshKernel/CurvilinearGridDeRefinement.cpp
```diff
@@ -0,0 +1,89 @@
...
+    for (auto m = 0, mIndexOriginalGrid = 0; mIndexOriginalGrid < m_grid->m_numM; ++m, mIndexOriginalGrid += localMDeRefinement)
+    {
+        localMDeRefinement = 1;
+        if (mIndexOriginalGrid >= mFirstNode && mIndexOriginalGrid < mSecondNode)
+        {
+            localMDeRefinement = numMToDeRefine;
+        }
+
+        deRefinedGrid.emplace_back(std::vector<Point>());
+        deRefinedGrid.back().reserve(m_grid->m_numN);
```
Hofer-Julian, line 74: Why specify back?

src/MeshKernel/CurvilinearGridDeRefinement.cpp
```diff
@@ -0,0 +1,89 @@
...
+        size_t localNDeRefinement;
+        for (auto n = 0, nIndexOriginalGrid = 0; nIndexOriginalGrid < m_grid->m_numN; ++n, nIndexOriginalGrid += localNDeRefinement)
+        {
+
+            localNDeRefinement = 1;
+            if (nIndexOriginalGrid >= nFirstNode && nIndexOriginalGrid < nSecondNode)
+            {
+                localNDeRefinement = numNToDeRefine;
+            }
+            deRefinedGrid.back().emplace_back(m_grid->m_gridNodes[mIndexOriginalGrid][nIndexOriginalGrid]);
```
Hofer-Julian, line 84: Why specify back

src/MeshKernelApi/MeshKernel.cpp
```diff
@@ -1634,7 +1634,7 @@ namespace meshkernelapi
         {
             if (meshKernelState.count(meshKernelId) == 0)
             {
-                throw std::invalid_argument("MeshKernel: The selected mesh kernel id does not exist.");
+                throw std::invalid_argument("mkernel_refine_curvilineargrid: The selected mesh kernel id does not exist.");
```
Hofer-Julian, line 1637: We currently have MeshKernel everywhere.
If we change this here, we should change it everywhere (I am not sure if its necessary though)

src/MeshKernelApi/MeshKernel.cpp
```diff
@@ -1660,6 +1660,42 @@ namespace meshkernelapi
...
+    MKERNEL_API int mkernel_derefine_curvilinear(int meshKernelId,
+                                                 const GeometryList& geometryListFirstPoint,
+                                                 const GeometryList& geometryListSecondPoint)
+    {
+        int exitCode = Success;
+        try
+        {
+            if (meshKernelState.count(meshKernelId) == 0)
+            {
+                throw std::invalid_argument("mkernel_derefine_curvilinear: The selected mesh kernel id does not exist.");
```
Hofer-Julian, line 1672: Use MeshKernel for now

src/MeshKernelApi/MeshKernel.cpp
```diff
@@ -1660,6 +1660,42 @@ namespace meshkernelapi
...
+            }
+            const auto firstPoint = ConvertGeometryListToPointVector(geometryListFirstPoint);
+
+            if (firstPoint.empty())
+            {
+                throw std::invalid_argument("mkernel_derefine_curvilinear: No first node of the segment defining the refinement zone has been provided.");
```
Hofer-Julian, line 1678: Same here

src/MeshKernelApi/MeshKernel.cpp
```diff
@@ -1660,6 +1660,42 @@ namespace meshkernelapi
...
+            }
+
+            const auto secondPoint = ConvertGeometryListToPointVector(geometryListSecondPoint);
+            if (secondPoint.empty())
+            {
+                throw std::invalid_argument("mkernel_derefine_curvilinear: No second node of the segment defining the refinement zone has been provided.");
```
Hofer-Julian, line 1684: Same here

src/MeshKernelApi/MeshKernel.cpp
```diff
@@ -1670,7 +1706,7 @@ namespace meshkernelapi
         {
             if (meshKernelState.count(meshKernelId) == 0)
             {
-                throw std::invalid_argument("MeshKernel: The selected mesh kernel id does not exist.");
+                throw std::invalid_argument("mkernel_compute_transfinite_from_splines_curvilinear: The selected mesh kernel id does not exist.");
```
Hofer-Julian, line 1709: Same here

Hofer-Julian (changes requested):
Looks good so far.
You can see my comments here

src/MeshKernel/CurvilinearGridDeRefinement.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+#include <utility>
+
+#include <MeshKernel/CurvilinearGrid.hpp>
+#include <MeshKernel/CurvilinearGridDeRefinement.hpp>
+#include <MeshKernel/Entities.hpp>
+#include <MeshKernel/Splines.hpp>
+
+meshkernel::CurvilinearGridDeRefinement::CurvilinearGridDeRefinement(std::shared_ptr<CurvilinearGrid> grid, const Point& firstPoint, const Point& secondPoint)
+    : m_grid(std::move(grid)),
```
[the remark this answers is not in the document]
lucacarniato: visual studio suggestion

src/MeshKernelApi/MeshKernel.cpp
lucacarniato: I think the error message will be more meaningful. let's say you have to debug this exception. By specifying the name of the function you know where to put the breakpoint exactly.

src/MeshKernel/CurvilinearGridDeRefinement.cpp
lucacarniato: Is not the same thing, the unit test will fail

src/MeshKernelApi/MeshKernel.cpp
Hofer-Julian: If I have access to a C++ compatible debugger, we don't need this information at all since we have access to the call stack.

The reason I added this prefix at all, is because I don't have that for languages like Python or Julia,

If I convert the exceptions to e.g. Python exceptions, I will add the name of the called function anyways so in that case the name would appear twice in the error message. That's why I'd prefer not to mention the name at the API level

src/MeshKernel/CurvilinearGridDeRefinement.cpp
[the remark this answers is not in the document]
Hofer-Julian: I guess, it doesn't hurt^^

src/MeshKernel/CurvilinearGridDeRefinement.cpp
Hofer-Julian: Ah yeah, max not min.
Replace it with 
```cpp
    const auto numMToDeRefine = std::max(1, mSecondNode - mFirstNode);
````
and everythings fine

src/MeshKernel/CurvilinearGridDeRefinement.cpp
lucacarniato: I am emplacing a vector at the back first, then for that vector at the back I reserve capacity. by doing so, I do not use indices

src/MeshKernel/CurvilinearGridDeRefinement.cpp
lucacarniato: See above

src/MeshKernel/CurvilinearGridDeRefinement.cpp
Hofer-Julian: Ah, now I understand the code!

Hofer-Julian (approved):

PR #83, opened 2021-03-22
GRIDEDIT-330: Curvilinear grid smoothing

lucacarniato:
…tion function for curvilinear grid, add OrthogonalProjectionOnSegment function on Operations


include/MeshKernel/CurvilinearGrid.hpp
```diff
@@ -54,6 +54,24 @@ namespace meshkernel
             Invalid        //(0)
         };
 
+        /// @brief A struct describing the column and row indices of a node
+        struct NodeIndices
+        {
+            /// @brief Overloads equality with another NodeIndices
+            bool operator==(const NodeIndices& rhs) const
```
luitjansl, line 61: if you use c++20 you can let the compiler make this operator
https://en.cppreference.com/w/cpp/language/default_comparisons

src/MeshKernel/CurvilinearGridDeRefinement.cpp
```diff
@@ -44,18 +44,18 @@ meshkernel::CurvilinearGridDeRefinement::CurvilinearGridDeRefinement(std::shared
...
 {
     // Get the m and n indices from the point coordinates
-    auto [mFirstNode, nFirstNode] = m_grid->GetNodeIndices(m_firstPoint);
-    auto [mSecondNode, nSecondNode] = m_grid->GetNodeIndices(m_secondPoint);
+    auto const firstNode = m_grid->GetNodeIndices(m_firstPoint);
+    auto const secondNode = m_grid->GetNodeIndices(m_secondPoint);
 
     /// The points must lie on the same gridline
-    if (mSecondNode - mFirstNode != 0 && nSecondNode - nFirstNode != 0)
+    if (secondNode.m - firstNode.m != 0 && secondNode.n - firstNode.n != 0)
```
luitjansl, line 51: now that you have a struct, can't  you make a function that checks this condition? that would be more expressive. 
say,
if (firstNode.IsSameGridline(SecondNode))

src/MeshKernel/CurvilinearGrid.cpp
```diff
@@ -136,7 +111,29 @@ std::tuple<std::vector<meshkernel::Point>, std::vector<meshkernel::Edge>, std::v
...
+    }
+    if (m_gridNodes[0].size() < 2)
+    {
+        return false;
+    }
+
+    return true;
+}
+
+meshkernel::CurvilinearGrid::NodeIndices meshkernel::CurvilinearGrid::GetNodeIndices(Point point)
```
luitjansl, line 136: you have here (and in other locations) really long typenames such as meshkernel::CurvilinearGrid::NodeIndices
You could make them smaller by using a "typedef" 
or alternatively put "using namespace meshkernel::CurvilinearGrid" at the top

lucacarniato: thanks, accounted for in the next commit

include/MeshKernel/CurvilinearGrid.hpp
lucacarniato: Using c++20 could be considered, needs to be investigated if it can be used in our TeamCity agents (namely an update of the boost libraries might be required)

PR #84, opened 2021-03-25
GRIDEDIT_327: line freeze implementation

lucacarniato:


src/MeshKernel/CurvilinearGridOrthogonalization.cpp
```diff
@@ -52,13 +49,14 @@ meshkernel::CurvilinearGridOrthogonalization::CurvilinearGridOrthogonalization(s
     m_d.resize(m_grid->m_numM, std::vector<double>(m_grid->m_numN, doubleMissingValue));
     m_e.resize(m_grid->m_numM, std::vector<double>(m_grid->m_numN, doubleMissingValue));
     m_atp.resize(m_grid->m_numM, std::vector<double>(m_grid->m_numN, doubleMissingValue));
+    m_isGridNodeFrozen.resize(m_grid->m_numM, std::vector<bool>(m_grid->m_numN, false));
```
luitjansl, line 52: here you are direcly accessing an attribute of another class (m_grid->m_numN) 
you could consider writig a getter function for it and use that instead: this reduces coupling between the classes

src/MeshKernel/CurvilinearGridOrthogonalization.cpp
```diff
@@ -67,19 +65,97 @@ void meshkernel::CurvilinearGridOrthogonalization::Compute()
...
+        auto const endCurrentLine = deltaMCurrentLine == 0 ? maxNCurrentLine : maxMCurrentLine;
+        auto const constantCoordinateCurrentLine = deltaMCurrentLine == 0 ? minMCurrentLine : minNCurrentLine;
+        for (auto i = startCurrentLine; i <= endCurrentLine; ++i)
+        {
+
+            for (auto j = startNewLine; j <= endNewLine; ++j)
+            {
+                if (j == constantCoordinateCurrentLine && i == constantCoordinateLine)
+                {
+                    throw AlgorithmError("CurvilinearGridOrthogonalization::SetFrozenLine the new line to freeze is crossing an existing line");
```
luitjansl, line 121: there is a preprocessor statement that returns the function name : 
cout << "Function name is: " << __FUNCTION__ << std::endl;
In any case, probably the stack trace tells you in which function the exception was raised, so I don't know if it is necesary to explicitely include it in the exception.

src/MeshKernel/CurvilinearGridOrthogonalization.cpp
lucacarniato: Hi, I understand the reason behind it. The approach of direct access of class members has been used elsewhere to avoid writing getter and setters for all class properties (which in this sort of numerical library are quite a lot). A change in the approach should be consistent across all files.

src/MeshKernel/CurvilinearGridOrthogonalization.cpp
```diff
@@ -67,19 +65,97 @@ void meshkernel::CurvilinearGridOrthogonalization::Compute()
...
+        throw std::invalid_argument("CurvilinearGridOrthogonalization::SetFrozenLine the points of the line to freeze must lie on the same grid line");
+    }
+
+    // The start-end coordinates of the new line, along m or n direction
+    auto const startNewLine = deltaMNewLine == 0 ? minN : minM;
+    auto const endNewLine = deltaMNewLine == 0 ? maxN : maxM;
+    auto const constantCoordinateLine = deltaMNewLine == 0 ? minM : minN;
+
+    // Frozen lines cannot cross existing frozen lines
+    for (auto const& frozenLine : m_frozenLines)
```
luitjansl, line 106: I'm not sure, but it looks like this for-loop is for the purpose of checking consistency, and not for changing the state of anything. If that is the case you could consider making a separate function for it to reduce the function complexity a bit

luitjansl (approved):

PR #85, opened 2021-04-06
GRIDEDIT-328: Add curvilinear grid line shift

lucacarniato:


include/MeshKernel/Constants.hpp
```diff
@@ -48,8 +48,8 @@ namespace meshkernel
     // geometric constants
     const double degrad_hp = M_PI / 180.0;                   ///< Conversion factor from degrees to radians(pi / 180)
     const double raddeg_hp = 180.0 / M_PI;                   ///< Conversion factor from radians to degrees(180 / pi)
-    const double earth_radius = 6378137.0;                   ///< Earth radius(m)
-    const double one_over_earth_radius = 1.0 / earth_radius; ///< One over earth_radius(m-1);
+    const double earth_radius = 6378137.0;                   ///< Earth radius(m_m)
```
luitjansl, line 51: m, not m_m

include/MeshKernel/Constants.hpp
```diff
@@ -74,7 +74,7 @@ namespace meshkernel
     const double mergingDistanceSquared = mergingDistance * mergingDistance; ///< Merging distance squared
 
     // physical constants
-    const double gravity = 9.81; ///< Gravitational acceleration on earth (m/s^2)
+    const double gravity = 9.81; ///< Gravitational acceleration on earth (m_m/s^2)
```
luitjansl, line 77: m, not m_m

include/MeshKernel/Contacts.hpp
```diff
@@ -61,8 +61,8 @@ namespace meshkernel
         /// The figure below shows two 2d meshes, a 1d mesh between them, and the 1d-2d contacts (in red).
         /// The boundary nodes of the 1d mesh (those sharing only one 1d edge) are not connected to any 2d face.
         /// For the 1d nodes not overlapping a 2d mesh,
-        /// a ray starting from the current node n is computed (dashed blue ray).
-        /// This ray is normal to the segment connecting the previous (n-1) and next one 1d node (n+1),
+        /// a ray starting from the current node m_n is computed (dashed blue ray).
```
luitjansl, line 64: this change could be unintentional

lucacarniato: yes, indeed. i used replacement, but ReSharper was not activated. a lesson I just learned

include/MeshKernel/CurvilinearGrid.hpp
```diff
@@ -73,64 +88,81 @@ namespace meshkernel
             /// @return True if on the same grid line, false otherwise
             bool IsOnTheSameGridLine(const NodeIndices& rhs) const
             {
-                if (m == rhs.m || n == rhs.n)
+                if (m_m == rhs.m_m || m_n == rhs.m_n)
```
luitjansl, line 91: you could make this a one liner:
return (m_M== ...)

src/MeshKernel/CurvilinearGridLineShift.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+using meshkernel::CurvilinearGrid;
+using meshkernel::CurvilinearGridLineShift;
+using meshkernel::Point;
+
+CurvilinearGridLineShift::CurvilinearGridLineShift(std::shared_ptr<CurvilinearGrid> grid) : CurvilinearGridAlgorithm(grid)
+
+{
+    // Compute the grid node types
+    m_gridModified = std::make_shared<CurvilinearGrid>(m_grid->m_gridNodes, m_grid->m_projection);
```
luitjansl, line 41: if your intention is to make a deep-copy of the input grid, you could create a copy or clone function in the CurviLinear grid class

src/MeshKernel/CurvilinearGridLineShift.cpp
```diff
@@ -0,0 +1,180 @@
...
+    auto previousDelta = m_gridModified->m_gridNodes[previousNodeIndex.m_m][previousNodeIndex.m_n] -
+                         m_grid->m_gridNodes[previousNodeIndex.m_m][previousNodeIndex.m_n];
+
+    const double eps = 1e-5;
+    auto previousCoordinate = m_lines[0].m_startCoordinate;
+    for (auto i = 1; i <= m_lines[0].m_endCoordinate; ++i)
+    {
+        auto const currentNodeIndex = m_lines[0].GetNodeindexFromCoordinate(i);
+
+        auto const currentDelta = m_gridModified->m_gridNodes[currentNodeIndex.m_m][currentNodeIndex.m_n] -
```
luitjansl, line 62: feel free to ignore this comment, I am in "stream of conciousness" mode. 
the type of currentDelta is "Point" but it could also be seen as a 2d vector. If it were a vector, then the check below could be made easier, because then you coul check currentdDelta.Norm() < eps instead of checking both components in turn.

include/MeshKernel/CurvilinearGridAlgorithm.hpp
```diff
@@ -0,0 +1,71 @@
...
+        /// @brief Sets a line, applicable to algorithms such as directional smoothing,
+        /// freeze a line in curvilinear grid orthogonalization and line shift
+        /// @param[in] firstPoint The point containing the first point of the line to freeze
+        /// @param[in] secondPoint The point containing the second point of the line to freeze
+        void SetLine(Point const& firstPoint, Point const& secondPoint);
+
+        /// @brief Virtual destructor
+        virtual ~CurvilinearGridAlgorithm() = default;
+
+        std::shared_ptr<CurvilinearGrid> m_grid;   ///< A pointer to the curvilinear grid to modify
```
luitjansl, line 66: If I'm not mistake you typically make a copy of the input grid with these algortihms and return the copy. If that is indeed the case, then you can make this attribute a std::shared_ptr<const CurvilinearGrid> so that the compiler will check you don't change it

include/MeshKernel/CurvilinearGridLine.hpp
```diff
@@ -0,0 +1,67 @@
...
+        [[nodiscard]] bool IsNodeOnLine(CurvilinearGrid::NodeIndices const& node) const;
+
+        /// @brief Gets the NodeIndices of a node on the grid line
+        /// @param[in] coordinate The one-dimensional coordinate along the grid line
+        /// @return The node indices
+        [[nodiscard]] CurvilinearGrid::NodeIndices GetNodeindexFromCoordinate(size_t const& coordinate) const;
+
+        CurvilinearGrid::NodeIndices m_startNode; ///<The start node of the grid line
+        CurvilinearGrid::NodeIndices m_endNode;   ///<The end node of the grid line
+        size_t m_startCoordinate;                 ///<The start coordinate. If it is an MGridLine, the start m otherwise the start n
```
luitjansl, line 62: I saw that you set these start coordinates in the constructor and then don't touch them anymore. Instead of having attributes, you could make functions that return these values. Then you have less state to manage.

include/MeshKernel/CurvilinearGridSmoothing.hpp
```diff
@@ -26,82 +26,54 @@
...
-                                       Point const& secondSegmentNode,
-                                       Point const& lowerLeftCornerSmoothingArea,
-                                       Point const& upperRightCornerSmootingArea);
-
-        /// @brief Sets the smoothing block
-        /// @param[in] firstCornerPoint The first point defining the smoothing area
-        /// @param[in] secondCornerPoint The second point defining the smoothing area
-        void SetBlock(Point const& firstCornerPoint, Point const& secondCornerPoint);
+        /// @return The smoothed grid
+        std::shared_ptr<CurvilinearGrid> Compute(Point const& lowerLeftCornerSmoothingArea,
```
luitjansl, line 57: this class has 2 methods called "compute" but in one of them you can specify a bounding box. If I hadn't read the doxygen comment I would have assumed that the second "compute" function does the same as the first one, but only within the bounding box.  But it turns out that it not the case. Maybe you could rename the compute function with the bounding box to something else?

include/MeshKernel/CurvilinearGridAlgorithm.hpp
```diff
@@ -0,0 +1,71 @@
...
+    /// @brief A class implementing the curvilinear grid orthogonalization algorithm
+    class CurvilinearGridAlgorithm
+    {
+    public:
+        /// @brief Class constructor
+        /// @param[in] grid                        The input curvilinear grid
+        CurvilinearGridAlgorithm(std::shared_ptr<CurvilinearGrid> grid);
+
+        /// @brief Executes the algorithm, pure virtual
+        virtual std::shared_ptr<CurvilinearGrid> Compute() = 0;
```
luitjansl, line 50: sometimes this function returns a copy of the input grid (like in CurvilinearGridLineShift), and sometimes it edits the input grid (lime in CurvilinearGridOrthogonalization)and returns the original  but modified input grid. That can be a bit confusing.

src/MeshKernel/CurvilinearGridFromSplines.cpp
```diff
@@ -283,7 +283,7 @@ void CurvilinearGridFromSplines::Initialize()
             continue;
         }
 
-        // construct the cross splines through the edges, along m discretization
+        // construct the cross splines through the edges, along m_m discretization
```
luitjansl, line 286: the same resharper issue :-)

src/MeshKernel/CurvilinearGridLineShift.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+        currentDelta = TransformDisplacement(currentDelta, currentNode, false);
+        m_gridModified->m_gridNodes[currentNode.m_m][currentNode.m_n] = m_grid->m_gridNodes[currentNode.m_m][currentNode.m_n] + currentDelta;
+    }
+}
+
+Point CurvilinearGridLineShift::TransformDisplacement(Point const& displacement, CurvilinearGrid::NodeIndices const& node, bool isLocal) const
+{
+    Point left = m_grid->m_gridNodes[node.m_m][node.m_n];
+    Point right = m_grid->m_gridNodes[node.m_m][node.m_n];
```
luitjansl, line 131: Point right = left ;
is more clear

src/MeshKernel/CurvilinearGridDeRefinement.cpp
```diff
@@ -87,6 +77,7 @@ CurvilinearGrid CurvilinearGridDeRefinement::Compute()
         }
         mIndexOriginalGrid += localMDeRefinement;
     }
+
     // substitute original grid with the derefined one
-    return CurvilinearGrid(std::move(deRefinedGrid), m_grid->m_projection);
-}
+    return std::make_shared<CurvilinearGrid>(std::move(deRefinedGrid), m_grid->m_projection);
```
luitjansl, line 82: be aware that because you use shared pointers everywhere, some instances could still keep the old curvilinear grid alive. maybe thats what you want, maybe not

include/MeshKernel/CurvilinearGridFromSplinesTransfinite.hpp
```diff
@@ -49,7 +49,7 @@ namespace meshkernel
 
         /// @brief Computes the adimensional intersections between splines.
```
luitjansl, line 50: same resharper issue here

include/MeshKernel/CurvilinearGridLine.hpp
lucacarniato: Nice catch

include/MeshKernel/CurvilinearGridAlgorithm.hpp
lucacarniato: Thanks. There are two approaches I think, One is conceiving the algorithms as modifiers of the grid instance (the approach i tried to follow). Another is to make a deep copy of the grid instance, where the algorithm can operate. In the end, this copy is returned from computing. Sometimes we need such a deep copy, that's why I struggled to have a uniform approach across all algorithm implementations. I now take the second approach, where a deep copy is made, assuming the cost of this deep copy is not too expensive. tests will tell

PR #95, opened 2021-05-11
GRIDEDIT 331: curvilinear line mirror

lucacarniato:


src/MeshKernel/CurvilinearGrid/CurvilinearGrid.cpp
```diff
@@ -506,91 +508,134 @@ void CurvilinearGrid::InsertFace(Point const& point)
...
+        gridSizeChanged = true;
     }
-    if (m_gridNodesTypes[firstNode.m_m][firstNode.m_n] == NodeType::Right ||
-        m_gridNodesTypes[secondNode.m_m][secondNode.m_n] == NodeType::Right)
+
+    return gridSizeChanged;
+}
+
+CurvilinearGrid::BoundaryGridLineType CurvilinearGrid::GetBoundaryGridLineType(NodeIndices const& firstNode, NodeIndices const& secondNode) const
+{
```
luitjansl, line 553: for safety, you could add ta check that the line between firstnnode and secondnode is actually a boundary line. Now it would return that the line is a BoundaryGridLineType::Bottom if firstnode is on the bottom and secondnode is internal, for example

src/MeshKernelApi/MeshKernel.cpp
```diff
@@ -2105,6 +2106,16 @@ namespace meshkernelapi
                 throw std::invalid_argument("MeshKernel: The selected mesh kernel state does not exist.");
             }
 
+            if (meshKernelState[meshKernelId].m_curvilinearGrid == nullptr)
```
luitjansl, line 2109: you could put these 2 checks in their own function, because they are done a lot

tests/unit/CurvilinearGridLineMirrorTests.cpp
```diff
@@ -0,0 +1,121 @@
+#include <gtest/gtest.h>
+
+#include <MeshKernel/CurvilinearGrid/CurvilinearGrid.hpp>
+#include <MeshKernel/CurvilinearGrid/CurvilinearGridLineMirror.hpp>
+#include <TestUtils/MakeCurvilinearGrids.hpp>
```
luitjansl, line 5: is there a test where the new gridline is not added (AddGridLineAtBoundary returning false)?

src/MeshKernel/CurvilinearGrid/CurvilinearGridLineMirror.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include <MeshKernel/CurvilinearGrid/CurvilinearGrid.hpp>
+#include <MeshKernel/CurvilinearGrid/CurvilinearGridLine.hpp>
+#include <MeshKernel/CurvilinearGrid/CurvilinearGridLineMirror.hpp>
+#include <MeshKernel/Entities.hpp>
+
+using meshkernel::CurvilinearGrid;
+using meshkernel::CurvilinearGridLineMirror;
+using meshkernel::Point;
+
+CurvilinearGridLineMirror::CurvilinearGridLineMirror(std::shared_ptr<CurvilinearGrid> grid, double mirroringFactor) : CurvilinearGridAlgorithm(grid), m_mirroringFactor(mirroringFactor)
```
luitjansl, line 37: can the mirroring factor be negative? would it create an internal gridline then?

luitjansl (approved):

PR #110, opened 2021-07-05
Correct contacts generation

lucacarniato:
Rename RTree::NearestNeighbors to RTree::NearestNeighbor
Add ComputeEdgesAndNodes method in MakeMeshes, for computing 1d/2d edges and nodes from legacy net files

### DIFF
--- a/include/MeshKernel/Mesh.hpp
+++ b/include/MeshKernel/Mesh.hpp
@@ -91,6 +91,13 @@ namespace meshkernel
             AdministrateMeshEdgesAndFaces
         };
 
+        /// Enumerator describing the different mesh types
+        enum class MeshTypes
+        {
+            Mesh1D,
+            Mesh2D
+        };
+
         /// @brief Default constructor
         Mesh() = default;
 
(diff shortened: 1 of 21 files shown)


include/MeshKernel/Mesh.hpp
```diff
@@ -91,6 +91,13 @@ namespace meshkernel
             AdministrateMeshEdgesAndFaces
         };
 
+        /// Enumerator describing the different mesh types
+        enum class MeshTypes
```
luitjansl, line 95: isn't this enum only used in testing code, such as MakeMeshes.* ? if so, you could define it there. Unless you foresee a future need for it...

src/MeshKernel/Mesh.cpp
```diff
@@ -637,22 +637,22 @@ void Mesh::BuildTree(MeshLocations meshLocation)
     }
```
luitjansl, line 637: in the function "buildtree", should you not check m_nodeRTreeRequiresUpdate before building the tree? now it could e the tree is not empty, but it still requires an update, and in that case it is not built. same for the edge tree, face tree.

src/MeshKernel/Mesh.cpp
```diff
@@ -637,22 +637,22 @@ void Mesh::BuildTree(MeshLocations meshLocation)
...
 }
 
-void Mesh::SearchNearestNeighbors(Point point, MeshLocations meshLocation)
+void Mesh::SearchNearestNeighbor(Point point, MeshLocations meshLocation)
 {
     BuildTree(meshLocation);
     if (meshLocation == MeshLocations::Nodes)
     {
-        m_nodesRTree.NearestNeighbors(point);
+        m_nodesRTree.NearestNeighbor(point);
```
luitjansl, line 645: I think you can reduce the number of if-statements here and in other functions if you make a function that returns the RTree given the meshlocation. Then this function would become 
{
     BuildTree(meshLocation);
     GetRTree(meshLocation).NearestNeighbor(point);
}

luitjansl (commented):
looks good!

PR #127, opened 2023-01-13
Feature/gridedit 418 make the compilers stricter

ahmads72:
Resolves GRIDEDIT-418, GRIDEDIT-414 and GRIDEDIT-441.
Please enable commit squashing if the repo is not configured as such.


tests/build_win/CMakeCache.txt
```diff
@@ -0,0 +1,338 @@
+# This is the CMakeCache file.
```
lucacarniato, line 1: we should not commit anything from the build folder

src/MeshKernelApi/MeshKernel.cpp
```diff
@@ -76,7 +76,7 @@ namespace meshkernelapi
     static int meshKernelStateCounter = 0;
 
     // Error state
-    static char exceptionMessage[512] = "";
+    static std::string exceptionMessage{};
```
lucacarniato, line 79: We have discussed this in the past. I believe std::string  has a char array that can be reallocated. having a static string does not guarantee that the underlying char array is also static (no reallocation of the underlying char array occurs). We want the exceptionMessage to be allocated in the same memory address during the execution of the library. This memory address is accessed by the client, and we do not want the client to access a memory containing something else.

tests/api/ApiTest.cpp
```diff
@@ -2572,7 +2572,7 @@ TEST_F(ApiTests, AveragingInterpolation_OnMesh2D_ShouldInterpolateValues)
 
     int const locationType = 1;          // Nodes
     int const averagingMethodType = 1;   // Simple averaging
-    int const relativeSearchSize = 1.01; // The relative search size
```
lucacarniato, line 2575: nicely spotted

tests/unit/Mesh2DTest.cpp
```diff
@@ -440,7 +440,8 @@ TEST(Mesh2D, GetNodeIndexShouldTriggerNodesRTreeBuild)
     ASSERT_EQ(0, mesh->m_nodesRTree.Size());
 
     // FindNodeCloseToAPoint builds m_nodesRTree for searching the nodes
-    const auto nodeIndex = mesh->FindNodeCloseToAPoint({1.5, 1.5}, 10.0);
+    const size_t index = mesh->FindNodeCloseToAPoint({1.5, 1.5}, 10.0);
+    ASSERT_TRUE(static_cast<long long>(index) >= 0); // Luca, need a better test here: ASSERT_EQ(index, actual_closest_node_index);
```
lucacarniato: corrected in my commit

tests/unit/Mesh2DTest.cpp
```diff
@@ -455,7 +456,8 @@ TEST(Mesh2D, FindEdgeCloseToAPointShouldTriggerEdgesRTreeBuild)
     auto mesh = MakeRectangularMeshForTesting(2, 2, 1.0, meshkernel::Projection::cartesian);
 
     // FindEdgeCloseToAPoint builds m_edgesRTree for searching the edges
-    const auto edgeIndex = mesh->FindEdgeCloseToAPoint({1.5, 1.5});
+    const size_t index = mesh->FindEdgeCloseToAPoint({1.5, 1.5});
+    ASSERT_TRUE(static_cast<long long>(index) >= 0); // Luca, need a better test here: ASSERT_EQ(index, actual_closest_edge_index);
```
lucacarniato: same here

lucacarniato (changes requested):

CMakeLists.txt
```diff
@@ -12,12 +12,13 @@ option(PRODUCE_CODE_COVERAGE "Produce code coverage files under GNU compilers" O
...
-    set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -g -O2 -std=c++17")
-    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g -std=c++17")
+    add_compile_options("-std=c++17;-Werror;-Wall;-Wextra;-pedantic;-Wno-unused-function")
+    add_compile_options("$<$<CONFIG:RELEASE>:-O2>")
+    add_compile_options("$<$<CONFIG:DEBUG>:-g>")
   else()
-    set(CMAKE_CXX_FLAGS_RELEASE
-        "${CMAKE_CXX_FLAGS_RELEASE} /EHsc /MP /std:c++17")
-    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /EHsc /MP /std:c++17")
+    add_compile_options("/EHsc;/MP;/std:c++17;/W3;/WX") #;/Wall")
```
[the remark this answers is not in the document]
ahmads72: Will do

src/MeshKernelApi/MeshKernel.cpp
ahmads72: I was trying to avoid the usage of strcpy, because it's very unsafe. Can you think of a better solution? memcpy comes to mind.

tests/api/ApiTest.cpp
ahmads72: It was the compiler :)

src/MeshKernelApi/MeshKernel.cpp
ahmads72: Used memcpy. This way the length of the source message cannot exceed the length of the destination. Up to 512 chars can now be copied from source to destination. If the length of the source is longer, the message is truncated.

lucacarniato (approved):

PR #129, opened 2023-01-17
Made public member vars private in class Polygons and provided const …

ahmads72:
Closes GRIDEDIT-420

### DIFF
--- a/src/MeshKernel/CurvilinearGrid/CurvilinearGridFromPolygon.cpp
+++ b/src/MeshKernel/CurvilinearGrid/CurvilinearGridFromPolygon.cpp
@@ -55,8 +55,7 @@ CurvilinearGrid CurvilinearGridFromPolygon::Compute(size_t firstNode,
     }
 
     // for the current polygon find the number of nodes
-    auto start = m_polygon->OuterIndices(0).first;
-    auto end = m_polygon->OuterIndices(0).second;
+    auto const& [start, end] = m_polygon->OuterIndices(0);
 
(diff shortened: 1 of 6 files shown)


src/MeshKernel/CurvilinearGrid/CurvilinearGridFromPolygon.cpp
```diff
@@ -55,8 +55,8 @@ CurvilinearGrid CurvilinearGridFromPolygon::Compute(size_t firstNode,
     }
 
     // for the current polygon find the number of nodes
-    auto start = m_polygon->m_outer_polygons_indices[0].first;
-    auto end = m_polygon->m_outer_polygons_indices[0].second;
+    auto start = m_polygon->OuterIndices(0).first;
```
lucacarniato, line 58: can use here structured bindings?

lucacarniato (changes requested):

src/MeshKernel/CurvilinearGrid/CurvilinearGridFromPolygon.cpp
ahmads72: Of course, I think we there are other places in this commit where it can be also used. I'll double check.

ahmads72: Turns out it's the only instance in this commit.

lucacarniato (approved):

PR #131, opened 2023-01-18
Feature/gridedit 416 declare proj consts as constexprs wih namespaces

ahmad-el-sayed:
Closes GRIDEDIT-416


tests/unit/ContactsTests.cpp
```diff
@@ -235,39 +235,39 @@ TEST(Contacts, ComputeContactsWithPolygons)
...
         {212.731567, 422.429504},
-        {meshkernel::doubleMissingValue, meshkernel::doubleMissingValue},
+        {meshkernel::constants::missing::doubleValue, meshkernel::constants::missing::doubleValue},
         {483.036316, 307.276855},
         {378.793213, 329.095245},
         {404.247925, 403.035400},
         {510.915283, 394.550476},
         {483.036316, 307.276855},
-        {meshkernel::doubleMissingValue, meshkernel::doubleMissingValue},
+        {meshkernel::constants::missing::doubleValue, meshkernel::constants::missing::doubleValue},
```
lucacarniato, line 251: using meshkernel::constants on the top of the .cpp file? so the namespace specification is shorter? maybe in the next pr

tests/unit/OrthogonalizationTests.cpp
```diff
@@ -355,12 +355,12 @@ TEST(OrthogonalizationAndSmoothing, OrthogonalizeAndSnapToLandBoundaries)
...
                                                 {429.349854, 450.959656},
                                                 {535.271545, 386.262909},
-                                                {meshkernel::doubleMissingValue, meshkernel::doubleMissingValue},
+                                                {meshkernel::constants::missing::doubleValue, meshkernel::constants::missing::doubleValue},
                                                 {246.995941, 262.285858},
                                                 {351.112183, 237.309906},
                                                 {443.191895, 262.285858},
                                                 {553.627319, 327.283539},
-                                                {meshkernel::doubleMissingValue, meshkernel::doubleMissingValue}};
+                                                {meshkernel::constants::missing::doubleValue, meshkernel::constants::missing::doubleValue}};
```
lucacarniato, line 363: meshkernel::constants::missing::doubleValue  could be reduced to missing::doubleValue

lucacarniato (changes requested):

include/MeshKernel/RTree.hpp
```diff
@@ -79,7 +79,7 @@ namespace meshkernel
 
             for (size_t n = 0; n < nodes.size(); ++n)
             {
-                if (nodes[n].x != doubleMissingValue && nodes[n].y != doubleMissingValue)
+                if (nodes[n].x != constants::missing::doubleValue && nodes[n].y != constants::missing::doubleValue)
```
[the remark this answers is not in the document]
ahmad-el-sayed: Looking at `IsEqual` doc, impl and intent,  I think it can be easily misused when the type is not integral. A lot can actually go wrong because it's a template function.  I suggest doing one of 2 things:

- restrict exclusively it to floating point types via the template parameter
- keep it as is but change the condition of the if statement to check for floating point types rather than integral types, in which case the operator == must be defined for the typename.


I'd go for option 1 as follows:
````
#include <iostream>
#include <limits>
#include <ios>
#include <type_traits>
#include <algorithm>
#include <string>

//template <std::floating_point T> // prefer this in this c++20, wat leuk!
template<typename T, typename std::enable_if<std::is_floating_point<T>::value>::type* = nullptr >
static bool IsEqual(const T value, T ref_value, T eps_mutilpier = 10.0)
{
    if (value == ref_value) 
    {
        return true;
    }
    else 
    {
        const T abs_diff = std::abs(value - ref_value);
        const T abs_value = std::abs(value);
        const T abs_ref_value = std::abs(ref_value);
        static const T tol = eps_mutilpier * std::numeric_limits<T>::epsilon();
        return abs_diff < tol * std::min(abs_value, abs_ref_value);
    }
}

int main() {
    {
        double const ref_value = 2.123456;
        double const value = ref_value + 5 * std::numeric_limits<double>::epsilon();
        std::cout << std::boolalpha << IsEqual(value, ref_value) << '\n'; // prints true
    }
    {
        float const ref_value = 2.12f;
        float const value = 2.1f;
        std::cout << std::boolalpha << IsEqual(value, ref_value) << '\n';// prints false
    }
    {
        std::string const ref_str("bla");
        std::string const str("blabla");
        // std::cout << std::boolalpha << IsEqual(value, ref_value) << '\n'; // Error
    }
    return 0;
}
````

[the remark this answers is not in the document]
ahmad-el-sayed: Went for option 2.

tests/unit/ContactsTests.cpp
ahmad-el-sayed: Done.

tests/unit/OrthogonalizationTests.cpp
ahmad-el-sayed: Done

lucacarniato (approved):

PR #142, opened 2023-04-05
Fix a bug on curvilinear grid generation using spherical coordinate 

lucacarniato:


ahmad-el-sayed:
@lucacarniato could you please fix the clang format check?

libs/MeshKernel/src/Mesh2D.cpp
```diff
@@ -37,6 +37,8 @@
 
 using meshkernel::Mesh2D;
```
ahmad-el-sayed, line 38: I'd simply put the class def in `namespace meshkernel { .... }`.
This applies to several classes throughout the repo.

libs/MeshKernel/include/MeshKernel/Mesh.hpp
```diff
@@ -122,6 +122,10 @@ namespace meshkernel
         /// @brief Default constructor
         Mesh() = default;
 
+        /// @brief  Constructs an empty mesh, sets only the projection
+        /// @param[in] projection  The projection to use
+        Mesh(Projection projection) : m_projection(projection){}
```
ahmad-el-sayed, line 127: Why not simply have a public method to set the projection in the base class Mesh? Something like
```
 Mesh::SetProjection(Projection projection) {  m_projection = projection; }
```
This saves you defining custom constructor in Mesh and all its derived classes: Mesh1D, Mesh2D and CurvilinearGrid.
Do you think this is possible?

lucacarniato: i think is better to set the projection in the constructor even if this requires a bit more code

libs/MeshKernel/src/Mesh2D.cpp
lucacarniato: I sis not use this form to avoid an extra level of nesting. But it could indeed be used

PR #159, opened 2023-05-03
correct intersection ordering

lucacarniato:


libs/MeshKernel/src/Mesh2D.cpp
```diff
@@ -1727,23 +1736,46 @@ Mesh2D::GetPolylineIntersections(const std::vector<Point>& polyLine)
...
-        std::stable_sort(edgesIntersections.begin(), edgesIntersections.end(),
-                         [](const EdgeMeshPolylineIntersection& first, const EdgeMeshPolylineIntersection& second)
-                         { return first.polylineSegmentDistance < second.polylineSegmentDistance; });
+        // edge intersections are unique
+        for (size_t e = 0; e < GetNumEdges(); ++e)
+        {
+            if (edgesIntersectionsResult[e].polylineDistance < 0)
+            {
+                edgesIntersectionsResult[e] = edgesIntersections[e];
+            }
```
BillSenior, line 1745: Can this be true?

ahmad-el-sayed (approved):

libs/MeshKernel/src/Mesh2D.cpp
lucacarniato: yes, because the result vector is filled with default initialized EdgeMeshPolylineIntersection, where polylineDistance is equal to  -999.0

PR #161, opened 2023-05-05
Simplify some unit test

lucacarniato:


libs/MeshKernelApi/tests/src/ApiTest.cpp
```diff
@@ -204,30 +204,16 @@ TEST_F(ApiTests, FlipEdges_WithALandBoundary_ShouldFlipEdges)
...
-        -0.5,
-        -0.5,
-        meshkernel::constants::missing::doubleValue});
-
-    std::unique_ptr<double> const zCoordinates(new double[4]{
-        0.0,
-        0.0,
-        0.0,
-        meshkernel::constants::missing::doubleValue});
+    std::vector xCoordinates{-0.5, -0.5, 4.0, meshkernel::constants::missing::doubleValue};
```
ahmad-el-sayed, line 207: Shall we not use template arg deduction here and be explicit about the typename?

ahmad-el-sayed (approved):

PR #169, opened 2023-06-02
GRIDEDIT-546 include wix installer files in repository

ScottWillcox:


package/msi/meshkernel_build_msi.cmd
ahmad-el-sayed, line 1: According to the the TC config, this is not used. Instead
- var wixDir is set in both candle and light steps 
- var name is hardcoded in the command lines

It would be actually nice to introduce a first step where the env vars are set and then used in candle and light, as done on this file.

ahmad-el-sayed (approved):

ahmad-el-sayed:
@ScottWillcox  should we cherry-pick this to release/v2.0.0? Otherwise TC job will fail when old config is removed.

PR #182, opened 2023-06-29
Feature/gridedit 588 change to bounded array

BillSenior:


libs/MeshKernel/src/Mesh2D.cpp
```diff
@@ -633,11 +638,13 @@ void Mesh2D::ComputeFaceClosedPolygon(size_t faceIndex, std::vector<Point>& poly
 {
     const auto numFaceNodes = GetNumFaceEdges(faceIndex);
     polygonNodesCache.clear();
-    polygonNodesCache.reserve(numFaceNodes);
+    polygonNodesCache.reserve(numFaceNodes + 1);
```
ahmad-el-sayed, line 641: good find

tools/test_utils/src/MakeMeshes.cpp
```diff
@@ -53,11 +55,11 @@ ReadLegacyMeshFile(std::filesystem::path const& file_path)
     }
 
     std::size_t num_edges;
-    nc_inq_dim(ncidp, dimid, read_name.get(), &num_edges);
-    std::shared_ptr<double> node_x(new double[num_nodes]);
-    std::shared_ptr<double> node_y(new double[num_nodes]);
-    std::shared_ptr<int> edge_nodes(new int[num_edges * 2]);
-    std::shared_ptr<int> edge_type(new int[num_edges]);
+    nc_inq_dim(ncidp, dimid, read_name.data(), &num_edges);
+    std::shared_ptr<double> node_x(new double[num_nodes], std::default_delete<double[]>());
```
ahmad-el-sayed, line 59: Can't we simply use vectors instead of shared_ptr? this requires using the data method instead of get in the calls nc_* below.

tools/test_utils/src/MakeMeshes.cpp
```diff
@@ -251,8 +253,8 @@ MakeRectangularMeshForApiTesting(
     auto num_x = num_columns + static_cast<size_t>(1);
 
     std::vector<std::vector<size_t>> indicesValues(num_x, std::vector<size_t>(num_y));
-    std::shared_ptr<double> node_x(new double[num_x * num_y]);
-    std::shared_ptr<double> node_y(new double[num_x * num_y]);
+    std::shared_ptr<double> node_x(new double[num_x * num_y], std::default_delete<double[]>());
```
ahmad-el-sayed, line 256: same

tools/test_utils/src/MakeMeshes.cpp
```diff
@@ -251,8 +253,8 @@ MakeRectangularMeshForApiTesting(
...
+    std::shared_ptr<double> node_y(new double[num_x * num_y], std::default_delete<double[]>());
```
ahmad-el-sayed, line 257: same

tools/test_utils/src/MakeMeshes.cpp
```diff
@@ -267,7 +269,7 @@ MakeRectangularMeshForApiTesting(
         }
     }
 
-    std::shared_ptr<int> edge_nodes(new int[((num_x - 1) * num_y + (num_y - 1) * num_x) * 2]);
+    std::shared_ptr<int> edge_nodes(new int[((num_x - 1) * num_y + (num_y - 1) * num_x) * 2], std::default_delete<int[]>());
```
ahmad-el-sayed, line 272: same

ahmad-el-sayed (changes requested):

tools/test_utils/src/MakeMeshes.cpp
BillSenior: Yes, I did think about that, but went for the least disruptive option.
I will change these to std::vector.

lucacarniato (approved):

tools/test_utils/src/MakeMeshes.cpp
lucacarniato: can we use make_shared, and avoid new?

lucacarniato (changes requested):

ahmad-el-sayed (approved):

PR #185, opened 2023-06-29
Feature/gridedit 590 pass point objs by const ref

ahmad-el-sayed:


libs/MeshKernel/include/MeshKernel/Exceptions.hpp
```diff
@@ -87,15 +88,22 @@ namespace meshkernel
...
+            : m_what(),
+              m_fmt_message(),
+              m_source_location(source_location)
+        {
+        }
 
         /// @brief Class constructor parametrized by a variadic error message and optionally the source location.
         /// @param[in] message         The variadic error message.
         /// @param[in] source_location The source location.
         MeshKernelError(VariadicErrorMessage const& message,
```
ahmad-el-sayed, line 103: Can any of you figure out why:
```
throw MeshKernelError({"some useful message id ={}", 5}); // (1) ambiguous
throw MeshKernelError(VariadicErrorMessage{"some useful message id ={}", 5)}; // (2) ok
throw MeshKernelError({"some useful message id ={} and value = {}", 5, 2.5}); // (3) ok
```
(1) compiler cannot pick a ctor when there is one variadic arg (see the other ctors in class)  
(2)  removes ambiguity  
(3) similar to (1), but has more than one variadic arg. Treatment done in (2) not required.

BillSenior: For me, I do no know what the problem is. Even after being in the language for >10 years, I have not used variadic templates, and std::format has only recently been added to gcc

BillSenior (approved):

libs/MeshKernelApi/include/MeshKernelApi/GeometryList.hpp
```diff
@@ -33,10 +33,10 @@ namespace meshkernelapi
     struct GeometryList
     {
         /// @brief The value used as separator in coordinates_x, coordinates_y and values
-        double geometry_separator;
+        double geometry_separator = -999.0;
```
BillSenior, line 36: Is this the same as constants::missing::doubleValue?

libs/MeshKernelApi/include/MeshKernelApi/GeometryList.hpp
```diff
@@ -33,10 +33,10 @@ namespace meshkernelapi
...
 
         /// @brief The value used to separate the inner part of a polygon from its outer part
-        double inner_outer_separator;
+        double inner_outer_separator = -998.0;
```
BillSenior, line 39: Is this the same as constants::missing::innerOuterSeparator?

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -63,26 +63,31 @@
...
+    static size_t const charBufferSize = 512;
+    static char exceptionMessage[charBufferSize] = "";
+    static meshkernel::MeshGeometryError meshGeometryError = meshkernel::MeshGeometryError(0, meshkernel::Mesh::Location::Unknown, "");
+
+    static void CopyExceptionMessage(const char* message)
+    {
+        size_t len = std::strlen(message);
+        size_t max_len = len > charBufferSize - 1 ? charBufferSize - 1 : len + 1;
+        std::memcpy(exceptionMessage, message, max_len);
+    }
```
BillSenior, line 89: What happens if message is nullptr?

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -95,33 +100,34 @@ namespace meshkernelapi
...
-            std::memcpy(exceptionMessage, e.what(), sizeof exceptionMessage);
+            CopyExceptionMessage(e.what());
             return MeshKernelApiErrors::NotImplemented;
         }
         catch (const meshkernel::MeshGeometryError& e)
         {
             meshGeometryError = e;
-            std::memcpy(exceptionMessage, e.what(), sizeof exceptionMessage);
+            CopyExceptionMessage(e.what());
             return MeshKernelApiErrors::MeshGeometryError;
```
BillSenior, line 110: the global meshGeometryError is only assigned in this catch block, should is be assigned in the other catch blocks too?

libs/MeshKernelApi/include/MeshKernelApi/GeometryList.hpp
lucacarniato: yes, let's default it to the separator used internally. the client can always get the correct value using the API available in meshkernel.

libs/MeshKernelApi/include/MeshKernelApi/GeometryList.hpp
lucacarniato: yes, let's default it to the separator used internally. same as above

PR #188, opened 2023-07-03
Introduce separate api for making uniform grids on extension

lucacarniato:
Start renaming api unit test names

### DIFF
--- a/libs/MeshKernelApi/src/MeshKernel.cpp
+++ b/libs/MeshKernelApi/src/MeshKernel.cpp
@@ -126,55 +126,6 @@ namespace meshkernelapi
         }
     }
 
-    meshkernel::CurvilinearGrid CreateUniformCurvilinearGrid(const meshkernel::MakeGridParameters& makeGridParameters,
-                                                             const GeometryList& geometryList,
-                                                             const meshkernel::Projection& projection)
-    {
-        meshkernel::CurvilinearGridCreateUniform curvilinearGridCreateUniform(projection);
-
-        auto polygonNodes = ConvertGeometryListToPointVector(geometryList);
-
-        const auto polygon = std::make_shared<meshkernel::Polygons>(polygonNodes, projection);
-
-        if (!polygon->IsEmpty())
-        {
-            return curvilinearGridCreateUniform.Compute(makeGridParameters.angle,
-                                                        makeGridParameters.block_size_x,
-                                                        makeGridParameters.block_size_y,
-                                                        polygon,
-                                                        0);
-        }
-        if (makeGridParameters.num_columns > 0 && makeGridParameters.num_rows > 0)
-        {
-            return curvilinearGridCreateUniform.Compute(makeGridParameters.num_columns,
-                                                        makeGridParameters.num_rows,
-                                                        makeGridParameters.origin_x,
-                                                        makeGridParameters.origin_y,
-                                                        makeGridParameters.angle,
-                                                        makeGridParameters.block_size_x,
-                                                        makeGridParameters.block_size_y);
-        }
-        throw meshkernel::AlgorithmError("The num_columns or num_rows in MakeGridParameters is not equal to 0");
-    }
-
-    meshkernel::CurvilinearGrid CreateUniformCurvilinearGridOnExtension(const meshkernel::MakeGridParameters& makeGridParameters,
-                                                                        const meshkernel::Projection& projection)
-    {
-        meshkernel::CurvilinearGridCreateUniform curvilinearGridCreateUniform(projection);
-
-        if (!meshkernel::IsEqual(makeGridParameters.angle, 0.0))
-        {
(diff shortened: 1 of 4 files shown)


libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -988,21 +988,42 @@ namespace meshkernelapi
             }
             else
             {
+                throw meshkernel::AlgorithmError("Invalid MakeGridParameters");
```
ahmad-el-sayed, line 991: We should use something more descriptive in the error message rather than the struct name.

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -988,21 +988,42 @@ namespace meshkernelapi
...
-                                                                                  makeGridParameters.origin_y,
-                                                                                  makeGridParameters.block_size_x,
-                                                                                  makeGridParameters.block_size_y,
-                                                                                  makeGridParameters.upper_right_x,
-                                                                                  makeGridParameters.upper_right_y);
-                auto const [nodes, edges, gridIndices] = curvilinearGrid.ConvertCurvilinearToNodesAndEdges();
-                *meshKernelState[meshKernelId].m_mesh2d += meshkernel::Mesh2D(edges, nodes, meshKernelState[meshKernelId].m_curvilinearGrid->m_projection);
+            meshkernel::CurvilinearGridCreateUniform curvilinearGridCreateUniform(meshKernelState[meshKernelId].m_projection);
+
+            if (!meshkernel::IsEqual(makeGridParameters.angle, 0.0))
```
ahmad-el-sayed, line 1014: We might as well perform range checks in mkernel_mesh2d_make_uniform

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -2333,18 +2354,40 @@ namespace meshkernelapi
             }
             else
             {
-                if (!meshkernel::IsEqual(makeGridParameters.angle, 0.0))
-                {
-                    throw meshkernel::AlgorithmError("When generating an uniform grid on an defined extension, the grid angle must be equal to 0");
-                }
+                throw meshkernel::AlgorithmError("Invalid MakeGridParameters");
```
ahmad-el-sayed, line 2357: Better error message

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -988,21 +988,42 @@ namespace meshkernelapi
...
+            exitCode = HandleExceptions(std::current_exception());
+        }
+        return exitCode;
+    }
 
-                if (!meshkernel::IsEqual(makeGridParameters.angle, 0.0))
-                {
-                    throw meshkernel::AlgorithmError("When generating an uniform grid on an defined extension, the grid angle must be equal to 0");
-                }
+    MKERNEL_API int mkernel_mesh2d_make_uniform_on_extension(int meshKernelId,
```
ahmad-el-sayed, line 1001: Check if mkernel_curvilinear_make_uniform_on_extension can be used here to duplication

ahmad-el-sayed (changes requested):

libs/MeshKernelApi/src/MeshKernel.cpp
lucacarniato: i removed the duplication by introducing a new function. Note that in the case of mesh2d_make_uniform_... we do not want to add a curvilinear grid in the meshkernel state, so calling mkernel_curvilinear_make_uniform_on_extension inside mesh2d_make_uniform_... will not work

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -126,6 +126,55 @@ namespace meshkernelapi
         }
     }
 
+    meshkernel::CurvilinearGrid CreateUniformCurvilinearGrid(const meshkernel::MakeGridParameters& makeGridParameters,
```
ahmad-el-sayed, line 129: Declare static or even better put in an anonymous namespace

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -126,6 +126,55 @@ namespace meshkernelapi
...
+                                                        makeGridParameters.origin_x,
+                                                        makeGridParameters.origin_y,
+                                                        makeGridParameters.angle,
+                                                        makeGridParameters.block_size_x,
+                                                        makeGridParameters.block_size_y);
+        }
+        throw meshkernel::AlgorithmError("The num_columns or num_rows in MakeGridParameters is not equal to 0");
+    }
+
+    meshkernel::CurvilinearGrid CreateUniformCurvilinearGridOnExtension(const meshkernel::MakeGridParameters& makeGridParameters,
```
ahmad-el-sayed, line 160: same

ahmad-el-sayed (changes requested):

PR #192, opened 2023-07-05
Feature/gridedit 510 polygon ref not working

BillSenior:


lucacarniato (approved):

libs/MeshKernel/include/MeshKernel/Entities.hpp
```diff
@@ -94,6 +94,39 @@ namespace meshkernel
         {
         }
 
+        /// @brief Inplace add point to point.
```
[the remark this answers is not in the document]
BillSenior: Isn't the spaceship operator about comparison?

libs/MeshKernel/include/MeshKernel/Entities.hpp
```diff
@@ -282,3 +310,59 @@ namespace meshkernel
     }
 
 } // namespace meshkernel
+
+inline meshkernel::Point& meshkernel::Point::operator+=(const Point& p)
```
ahmad-el-sayed, line 314: Definitions can be moved to the class declaration where the other operators are overloaded.   
Isn't inlining in this case done automatically by the compiler?

libs/MeshKernel/src/Polygons.cpp
```diff
@@ -145,11 +146,14 @@ std::vector<meshkernel::Point> Polygons::RefineFirstPolygon(size_t startIndex,
 
     if (endIndex <= startIndex)
     {
-        throw std::invalid_argument("Polygons::RefineFirstPolygon: The end index is smaller than the start index.");
+        throw std::invalid_argument("Polygons::RefineFirstPolygon: The end index is smaller than the start index: " + std::to_string(startIndex) + " >= " + std::to_string(endIndex));
```
ahmad-el-sayed, line 149: You can use out custom exceptions here, See Exceptions.hpp.

libs/MeshKernel/src/Polygons.cpp
```diff
@@ -163,88 +167,61 @@ std::vector<meshkernel::Point> Polygons::RefineFirstPolygon(size_t startIndex,
 
     if (!areIndicesValid)
     {
-        throw std::invalid_argument("Polygons::RefineFirstPolygon: The indices are not valid.");
+        throw std::invalid_argument("Polygons::RefineFirstPolygon: The indices are not valid: " + std::to_string(startIndex) + ", " + std::to_string(endIndex));
```
ahmad-el-sayed, line 170: same

libs/MeshKernel/src/Polygons.cpp
```diff
@@ -163,88 +167,61 @@ std::vector<meshkernel::Point> Polygons::RefineFirstPolygon(size_t startIndex,
...
-            if (!snappedToLastPoint)
-            {
-                refinedPolygon.emplace_back(m_nodes[nextNodeIndex]);
-            }
+        Point p = m_nodes[i];
+        Point delta = m_nodes[i + 1] - m_nodes[i];
+        double segmentLength = ComputeDistance(m_nodes[i], m_nodes[i + 1], m_projection);
+        double lengthAlongInterval = refinementDistance;
+        // TODO how large should the tolerance be? 1.0e-3 seems quite tight for this algorithm
+        double tolerance = 1.0e-3;
```
ahmad-el-sayed, line 208: Constants.hpp is a better suited place for constants. Make constexpr.

libs/MeshKernel/src/Polygons.cpp
```diff
@@ -163,88 +167,61 @@ std::vector<meshkernel::Point> Polygons::RefineFirstPolygon(size_t startIndex,
...
-        if (pointLengthCoordinate > nodeLengthCoordinate[nextNodeIndex])
-        {
-            // if not snapped to the original last polygon point, snap it
-            if (!snappedToLastPoint)
-            {
-                refinedPolygon.emplace_back(m_nodes[nextNodeIndex]);
-            }
+        Point p = m_nodes[i];
+        Point delta = m_nodes[i + 1] - m_nodes[i];
+        double segmentLength = ComputeDistance(m_nodes[i], m_nodes[i + 1], m_projection);
```
ahmad-el-sayed, line 205: Make const

libs/MeshKernel/tests/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ set(
   ${SRC_DIR}/MeshRefinementTests.cpp
   ${SRC_DIR}/MeshTests.cpp
   ${SRC_DIR}/OrthogonalizationTests.cpp
+  ${SRC_DIR}/PointTests.cpp
```
ahmad-el-sayed, line 35: thanks for preserving the alphabetical order :)

libs/MeshKernel/tests/src/PointTests.cpp
ahmad-el-sayed, line 1: It would be great if you can extend the tests to include the existing operators.

libs/MeshKernel/tests/src/PolygonTests.cpp
```diff
@@ -122,6 +147,171 @@ TEST(Polygons, RefinePolygon)
...
+
+    meshkernel::Polygons polygons(nodes, meshkernel::Projection::cartesian);
+
+    // Execute
+    const auto refinedPolygon = polygons.RefineFirstPolygon(0, 0, 1.0);
+
+    meshkernel::Polygons polygons2(refinedPolygon, meshkernel::Projection::cartesian);
+    const auto refinedPolygon2 = polygons2.RefineFirstPolygon(0, 0, 1.0);
+
+    const double tolerance = 1e-5;
```
ahmad-el-sayed, line 169: constexpr

libs/MeshKernel/tests/src/PolygonTests.cpp
```diff
@@ -122,6 +147,171 @@ TEST(Polygons, RefinePolygon)
...
+
+    meshkernel::Polygons polygons(nodes, meshkernel::Projection::cartesian);
+
+    // Execute
+    const auto refinedPolygon = polygons.RefineFirstPolygon(0, 0, 1.0);
+
+    meshkernel::Polygons polygons2(refinedPolygon, meshkernel::Projection::cartesian);
+    const auto refinedPolygon2 = polygons2.RefineFirstPolygon(0, 0, 2.0);
+
+    const double tolerance = 1e-13;
```
ahmad-el-sayed, line 223: constexpr

libs/MeshKernel/tests/src/PolygonTests.cpp
```diff
@@ -122,6 +147,171 @@ TEST(Polygons, RefinePolygon)
...
+
+    meshkernel::Polygons polygons(nodes, meshkernel::Projection::cartesian);
+
+    // Execute
+    const auto refinedPolygon = polygons.RefineFirstPolygon(0, 0, 1.0);
+
+    meshkernel::Polygons polygons2(refinedPolygon, meshkernel::Projection::cartesian);
+    const auto refinedPolygon2 = polygons2.RefineFirstPolygon(0, 0, 0.5);
+
+    const double tolerance = 1e-5;
```
ahmad-el-sayed, line 256: constexpr

libs/MeshKernel/tests/src/PolygonTests.cpp
```diff
@@ -158,6 +348,112 @@ TEST(Polygons, RefinePolygonOneSide)
...
+
+    // Execute
+    const auto refinedPolygon = polygons.RefineFirstPolygon(0, 1, 1.0);
+
+    meshkernel::Polygons polygons2(refinedPolygon, meshkernel::Projection::cartesian);
+    // Now there are additional segments in the polygon, the segments to refine is different.
+    const auto refinedPolygon2 = polygons2.RefineFirstPolygon(0, 3, 1.0);
+
+    ASSERT_EQ(7, refinedPolygon.size());
+    const double tolerance = 1e-10;
```
ahmad-el-sayed, line 372: constexpr

libs/MeshKernel/tests/src/PolygonTests.cpp
```diff
@@ -158,6 +348,112 @@ TEST(Polygons, RefinePolygonOneSide)
...
+
+    // Execute
+    const auto refinedPolygon = polygons.RefineFirstPolygon(1, 2, 2.0);
+
+    meshkernel::Polygons polygons2(refinedPolygon, meshkernel::Projection::cartesian);
+
+    // Refine the same edge, this time there should be two segments making up the edge.
+    const auto refinedPolygon2 = polygons2.RefineFirstPolygon(1, 3, 1.0);
+
+    const double tolerance = 1.0e-13;
```
ahmad-el-sayed, line 420: constexpr

tools/test_utils/src/MakeMeshes.cpp
```diff
@@ -53,11 +55,11 @@ ReadLegacyMeshFile(std::filesystem::path const& file_path)
     }
 
     std::size_t num_edges;
-    nc_inq_dim(ncidp, dimid, read_name.get(), &num_edges);
-    std::shared_ptr<double> node_x(new double[num_nodes]);
-    std::shared_ptr<double> node_y(new double[num_nodes]);
-    std::shared_ptr<int> edge_nodes(new int[num_edges * 2]);
-    std::shared_ptr<int> edge_type(new int[num_edges]);
+    nc_inq_dim(ncidp, dimid, read_name.data(), &num_edges);
+    std::shared_ptr<double> node_x(new double[num_nodes], std::default_delete<double[]>());
```
ahmad-el-sayed, line 59: Did we change all those unique_ptrs to vectors in the release/v2.0.0 branch? If yes, then never mind. When the release branch is merged, this will be taken care off. Otherwise Please make the change here.

tools/test_utils/src/MakeMeshes.cpp
```diff
@@ -251,8 +253,8 @@ MakeRectangularMeshForApiTesting(
     auto num_x = num_columns + static_cast<size_t>(1);
```
ahmad-el-sayed, line 253: hmmm :)

tools/test_utils/src/SampleFileReader.cpp
```diff
@@ -1,5 +1,6 @@
 #include "SampleFileReader.hpp"
 
+#include <algorithm>
```
ahmad-el-sayed, line 3: Is GCC 13.1 complaining? We should bump the version on the build server.

ahmad-el-sayed (changes requested):

libs/MeshKernel/src/Polygons.cpp
BillSenior: I created a new exception ConstraintError and this will be raised for these constraint violations. It can (should) also be used in other constraint violation cases.

libs/MeshKernel/src/Polygons.cpp
BillSenior: Fixed.

tools/test_utils/src/MakeMeshes.cpp
BillSenior: I think that I changed them all in the release branch

tools/test_utils/src/SampleFileReader.cpp
BillSenior: This is also changed in the release branch

BillSenior: Do you think we should use the clang compiler on mac os? 
 I think it is the default compiler for this platform.
Then we also get the code compiled with another compiler.

tools/test_utils/src/MakeMeshes.cpp
BillSenior: No idea why this is like this.
I have changed it.

libs/MeshKernel/tests/src/PolygonTests.cpp
BillSenior: fixed

libs/MeshKernel/tests/src/PolygonTests.cpp
BillSenior: fixed

libs/MeshKernel/tests/src/PolygonTests.cpp
BillSenior: fixed

libs/MeshKernel/tests/src/PolygonTests.cpp
BillSenior: fixed

libs/MeshKernel/include/MeshKernel/Entities.hpp
BillSenior: Like we discussed, the Entities file needs to be refactored.

libs/MeshKernel/src/Polygons.cpp
BillSenior: as above, now uses ConstraintError

libs/MeshKernel/tests/src/PolygonTests.cpp
BillSenior: fixed

libs/MeshKernel/src/Polygons.cpp
BillSenior: Done

libs/MeshKernel/tests/src/PointTests.cpp
BillSenior: Done

libs/MeshKernel/src/Polygons.cpp
ahmad-el-sayed: Ok

tools/test_utils/src/SampleFileReader.cpp
ahmad-el-sayed: On local branch yes. Had to fix various issues that GCC overlooks. macOS build uses GCC as well.

libs/MeshKernel/include/MeshKernel/Constants.hpp
```diff
@@ -67,6 +67,7 @@ namespace meshkernel
             constexpr double earth_radius = 6378137.0;                  ///< Earth radius(m)
             constexpr double inverse_earth_radius = 1.0 / earth_radius; ///< One over constants::geometric::earth_radius(m-1);
             constexpr double absLatitudeAtPoles = 0.0001;               ///< Pole tolerance in degrees
+            constexpr double refinementTolerance = 1.0e-2;              ///< Relative size of refinement.
```
ahmad-el-sayed, line 70: Before addressing this comment, you used `double tolerance = 1.0e-3;`. This became here 1.0e-2. What happened?

BillSenior: When refining the polygons, each polyline segment is refined so that it has refined segments of x-metres. The tolerance is used to control the size of the smallest refined segment.

Example, the original segment length is 1001 metres, the user wants to have a maximum segment length of 100 metres, the algorithm will divide the original segment into 10 segments of 100 metres and 1 of 1 metre. This is the limit of the tolerance. If the original lenght was 1000.9 then there would be 9 segments of 100 metres and 1 of 100.9 metres.

I have two issues with the algorithm as it currently stands:
1. I think that 1.0e-2 is also too big, probably 0.1 might be a better choice
2. better still would be to have an option for a uniform refinement, but this is different from the original algorithm or would require an API change (an extra parameter indicating a uniform spacing is required)

ahmad-el-sayed (approved):

PR #197, opened 2023-07-11
Added Eigen dependency (GRIDEDIT-602)

ahmad-el-sayed:


cmake/fetch_content.cmake
```diff
@@ -83,4 +83,20 @@ if(${USE_LIBFMT})
...
-  
\ No newline at end of file
+  
+# Eigen
+# Note: v3.4.0 seems to have a problem detecting c++11 when MSVC is used, so master will be used here.
+FetchContent_Declare(
+  Eigen
+  GIT_REPOSITORY https://gitlab.com/libeigen/eigen.git
+  GIT_TAG        master
+)
```
BillSenior, line 93: Could we have unit test failures when somebody makes a commit to the Eigen library if we always the head of the master?
Would it be an idea to use a specific commit to the master?

BillSenior: On a previous project somebody made a change to the testing environment that it always used the latest version of the Intel c++ compiler (why not, seems like a good idea)(we did not do much in the way of code reviewing at the time). Then one day, we had failures in many of the tests. We spent a a long time looking for the problem, undoing work getting really confused why we had these failures. We eventually found the cause, a system admin somewhere (not in our company) made an update to the Intel compiler.

libs/MeshKernel/include/MeshKernel/Entities.hpp
```diff
@@ -159,7 +161,17 @@ namespace meshkernel
...
+
+    /// @brief Row major dynamic matrix
+    /// @tparam T Data type
+    template <class T>
+    using MatrixRowMajor = Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>;
+
+    /// @brief Column major dynamic matrix
+    /// @tparam T Data type
+    template <class T>
+    using MatrixColMajor = Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic, Eigen::ColMajor>;
```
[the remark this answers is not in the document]
BillSenior: Might consider a nested namespace, e.g. linalg?

cmake/fetch_content.cmake
ahmad-el-sayed: That's a very good point. I'll check out a commit in a working state instead.

libs/MeshKernel/include/MeshKernel/Entities.hpp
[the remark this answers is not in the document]
ahmad-el-sayed: Sure can. The project in general lacks namespaces.

cmake/fetch_content.cmake
ahmad-el-sayed: Done

libs/MeshKernel/include/MeshKernel/Entities.hpp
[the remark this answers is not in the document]
ahmad-el-sayed: This has been moved to the dedicated header `libs/MeshKernel/include/MeshKernel/Utilities/LinearAlgebra.hpp`

BillSenior (approved):
Looks good

PR #207, opened 2023-07-28
Feature/gridedit 611 snap polygon to land boundary

BillSenior:


libs/MeshKernel/include/MeshKernel/Polygons.hpp
```diff
@@ -63,6 +67,12 @@ namespace meshkernel
         /// @return The new offset polygon
         [[nodiscard]] Polygons OffsetCopy(double distance, bool innerAndOuter) const;
 
+        /// @brief Snap the polygon to the land boundary
+        ///
+        /// The polygon points are snapped to the closest point on the land boundary.
+        /// @param[in] landBoundary The land boundary to which the polygon should be snapped.
+        void SnapToLandBoundary(const LandBoundary& landBoundary);
```
lucacarniato, line 74: I think we also need to select the polygon indices for selecting the part of the polygon that will be snapped

lucacarniato (changes requested):

libs/MeshKernel/include/MeshKernel/Polygons.hpp
BillSenior: Added start and end index

libs/MeshKernel/src/LandBoundary.cpp
```diff
@@ -12,29 +12,44 @@ void meshkernel::LandBoundary::FindNearestPoint(const Point& samplePoint,
...
     nearestPoint = samplePoint;
-    segmentStartIndex = constants::missing::intValue;
+    segmentStartIndex = constants::missing::uintValue;
     scaledDistanceToStart = -1.0;
 
-    // TODO better value for initial value and move to constants namespace
     minimumDistance = 9.0e33;
 
-    for (UInt i = 0; i < m_nodes.size() - 1; ++i)
+    if (samplePoint.IsValid())
```
lucacarniato, line 20: to avoid nesting you could return here?

lucacarniato (changes requested):

libs/MeshKernel/src/LandBoundary.cpp
BillSenior: I think here the comparison tool in github has made a mistake

BillSenior: If  you are comparing the code 

21 for (Uint i = 0 ...)

to

20 if(samplePoint.IsValid

BillSenior: now I see what you mean

BillSenior: changed

lucacarniato (approved):

PR #208, opened 2023-07-28
GRIDEDIT-617 Added spline snapping to api

BillSenior:


libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -1710,6 +1711,106 @@ namespace meshkernelapi
...
+
+            //--------------------------------
+            // Copy points from parameters to create land boundary and splines
+
+            for (int i = 0; i < land.num_coordinates; ++i)
+            {
+                landBoundaryPoints[i] = meshkernel::Point({land.coordinates_x[i], land.coordinates_y[i]});
+            }
+
+            for (int i = 0; i < splines.num_coordinates; ++i)
```
lucacarniato, line 1765: what if multiple splines are provided? start/end index refer to the selected spline corner's point

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -1710,6 +1711,106 @@ namespace meshkernelapi
...
+            {
+                splinePoints[i] = meshkernel::Point({splines.coordinates_x[i], splines.coordinates_y[i]});
+            }
+
+            meshkernel::LandBoundary landBoundary(landBoundaryPoints);
+            meshkernel::Splines splineValues(meshKernelState[meshKernelId].m_mesh2d->m_projection);
+
+            splineValues.AddSpline(splinePoints, startSplineIndex, static_cast<meshkernel::UInt>(splinePoints.size()));
```
lucacarniato, line 1773: suppose we pass to the API only the spline that needs to be snapped in a set of several splines, the startSplineIndex and endSplineIndex would not be required. what we need instead are the indeces of the the corner points defining the portion of the spline that will be snapped

lucacarniato (changes requested):

lucacarniato (approved):

PR #209, opened 2023-08-01
Feature/gridedit 616 api function span polygon to land

BillSenior:


libs/MeshKernelApi/include/MeshKernelApi/MeshKernel.hpp
```diff
@@ -408,6 +408,16 @@ namespace meshkernelapi
                                                              int startSplineIndex,
                                                              int endSplineIndex);
 
+        /// @brief Snaps the polygon to the land boundary
+        ///
+        /// @param[in] meshKernelId     The id of the mesh state
+        /// @param[in] land             The land boundary
+        /// @param[in] polygon          The polygon values to be snapped
+        /// @returns Error code
+        MKERNEL_API int mkernel_polygon_snap_to_landboundary(int meshKernelId,
```
lucacarniato, line 417: should we also pass the indices of the polygon vertices defining the portion of the polygon that should be snapped?

lucacarniato (changes requested):

libs/MeshKernelApi/include/MeshKernelApi/MeshKernel.hpp
BillSenior: Added start and end index

lucacarniato (approved):

PR #224, opened 2023-09-19
Feature/gridedit 652 range checks

ahmad-el-sayed:


lucacarniato (changes requested):

libs/MeshKernel/include/MeshKernel/Formatting.hpp
```diff
@@ -0,0 +1,82 @@
...
+    /// @tparam FormatContext
+    /// @param vector The vector to be formatted
+    /// @param format_context The format context
+    /// @return Iterator past the end of the context's output buffer
+    template <typename FormatContext>
+    auto format(std::vector<T> const& vector,
+                FormatContext& format_context) const
+    {
+        // get ref to the output buffer
+        auto&& out = format_context.out();
```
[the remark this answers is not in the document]
ahmad-el-sayed: Cannot be const, `dts::format_to` writes to `out`.

libs/MeshKernel/include/MeshKernel/RangeCheck.hpp
```diff
@@ -0,0 +1,501 @@
...
+        inline static std::unordered_map<Comparison, std::string> const ValidRangeFormat = {
+            {Comparison::Equal, "{} = {}"},
+            {Comparison::NotEqual, "{} != {}"},
+            {Comparison::Greater, "{} > {}"},
+            {Comparison::GreaterEqual, "{} >= {}"},
+            {Comparison::Less, "{} < {}"},
+            {Comparison::LessEqual, "{} <= {}"},
+            {Comparison::InClosedInterval, "{} <= {} <= {}"},
+            {Comparison::InOpenInterval, "{} <= {} <= {}"},
+            {Comparison::InSemiOpenFromAboveInterval, "{} <= {} < {}"},
```
BillSenior, line 83: Incorrect comparison operator for open interval

libs/MeshKernel/include/MeshKernel/RangeCheck.hpp
```diff
@@ -0,0 +1,501 @@
...
+            {Comparison::InSemiOpenFromBelowInterval, "{} < {} <= {}"},
```
BillSenior, line 84: (Not necessary to do)
Perhaps change name to InRightHalfOpenInterval and inLeftHalfOpenInterval

libs/MeshKernel/include/MeshKernel/RangeCheck.hpp
```diff
@@ -0,0 +1,501 @@
...
+                    interval_upper_bound);
+            }
+        }
+
+        /// @brief Checks the validity of a value given a vector of values, supports the predicates one of and none of
+        /// @tparam T The type of the value to check (RangeCheckableType)
+        /// @param value The value to check
+        /// @param values The vector of values
+        /// @param predicate The comparison predicate (one of <= ... <=, < ... <, < ... <=, and <= ... <)
+        /// @param comparison The key of the performed comparison
```
BillSenior, line 149: Incorrect predicate comment

BillSenior: One question about the IsOneOf and IsNoneOf.
It probably will not occur: what if floating point numbers are being checked and one of the values, either the value to check or the set of number it will be checked against is not a literal but the result of some calculation. Can you be sure that you will get strict equality?

libs/MeshKernel/src/CurvilinearGrid/CurvilinearGridCreateUniform.cpp
```diff
@@ -256,25 +225,16 @@ namespace meshkernel
...
-        {
-            throw AlgorithmError("BlockSizeX cannot be <= 0");
-        }
-        if (blockSizeY <= 0.0)
-        {
-            throw AlgorithmError("BlockSizeY cannot be <= 0");
-        }
+
+        range_check::CheckInOpenInterval(angle, {-90.0, 90.0}, "Grid angle");
+        range_check::CheckGreater(blockSizeX, 0.0, "X block size");
```
BillSenior, line 230: Should the angle be checked that it is in the closed interval?
The code it replaces was for the closed interval (std::abs(angle) > 90.0).
However, the comment in the original does match your code.

I think it should be either half open or a closed interval

libs/MeshKernel/tests/src/RangeCheckTests.cpp
```diff
@@ -0,0 +1,170 @@
...
+    {
+        EXPECT_NO_THROW(range_check::CheckEqual<T>(1, 1, name));
+        EXPECT_THROW(range_check::CheckEqual<T>(1, 2, name), RangeError);
+
+        if (std::same_as<T, float> ||
+            std::same_as<T, double> ||
+            std::same_as<T, long double>)
+        {
+            double constexpr x = 1.0;
+            double constexpr epsilon = std::numeric_limits<T>::epsilon();
```
BillSenior, line 26: May be better with the "if" changed to "if constexpr"

BillSenior, line 26: Replace double with T, otherwise only testing for double

Very minor (even less than very minor), since long double is there what about std::float128_t (not sure if this is c++23 only)(equivalent to __float128)

BillSenior, line 26: When changing from double to T, may need a static cast from 1.0 to type T

libs/MeshKernel/include/MeshKernel/Exceptions.hpp
```diff
@@ -57,15 +49,16 @@ namespace meshkernel
...
-        StdLibExceptionCode = 7,    ///< Standrad library exception
-        UnknownExceptionCode = 8    ///< Unknown exception
+        Success = 0,            ///< Success
+        MeshKernelErrorCode,    ///< MehKernel error
+        NotImplementedCode,     ///< Not implemented error
+        AlgorithmErrorCode,     ///< Algorithm error
+        ConstraintErrorCode,    ///< Constraint error
+        MeshGeometryErrorCode,  ///< Geometry error
+        LinearAlgebraErrorCode, ///< Lienar algebra error
+        RangeErrorCode,         ///< Range error
```
BillSenior, line 59: (existing) Spelling error on comment

libs/MeshKernel/include/MeshKernel/RangeCheck.hpp
ahmad-el-sayed: Had a hard time choosing something suitable. Will change.

libs/MeshKernel/include/MeshKernel/RangeCheck.hpp
ahmad-el-sayed: Does the unit test answer your question?

libs/MeshKernel/include/MeshKernel/RangeCheck.hpp
ahmad-el-sayed: Done

libs/MeshKernel/tests/src/RangeCheckTests.cpp
ahmad-el-sayed: Yes, saw thiose, in fact this->CheckEqual() and this->CheckNotEqual() were never called in the typed test.

[std::is_floating_point](https://en.cppreference.com/w/cpp/types/is_floating_point) , which is used in the concept std::floating_point, covers std::float128_t (sice c++23). I did not check the outcome of the test though.

libs/MeshKernel/include/MeshKernel/Exceptions.hpp
ahmad-el-sayed: Fixed

libs/MeshKernel/src/CurvilinearGrid/CurvilinearGridCreateUniform.cpp
ahmad-el-sayed: @BillSenior `if (std::abs(angle) > 90.0)` is equiv to `if (angle < -90.0 || angle  > 90.0)`. So we want what's in between, -90 <= angle <=90, therefore `CheckInClosedInterval` must be used rather than `CheckInOpenInterval`.

libs/MeshKernel/include/MeshKernel/Exceptions.hpp
```diff
@@ -329,4 +342,20 @@ namespace meshkernel
...
+    public:
+        /// @brief Class constructor
+        using MeshKernelError::MeshKernelError;
+
+    private:
+        /// @brief Returns the error category.
+        /// @return The  error category.
+        [[nodiscard]] ErrorCategory Category() const override
+        {
+            return {"RangeError", ExitCode::RangeErrorCode};
```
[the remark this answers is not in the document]
ahmad-el-sayed: This is allowed since c++11.

BillSenior (approved):

PR #232, opened 2023-10-02
Feature/gridedit 665 performance of face area centre of mass

BillSenior:


libs/MeshKernel/src/Polygon.cpp
```diff
@@ -349,21 +349,78 @@ std::tuple<double, meshkernel::Point, meshkernel::TraversalDirection> meshkernel
     const Point reference = ReferencePoint(polygon, projection);
     const auto numberOfPointsOpenedPolygon = static_cast<UInt>(polygon.size()) - 1;
 
-    for (UInt n = 0; n < numberOfPointsOpenedPolygon; ++n)
+    if (numberOfPointsOpenedPolygon == 3)
```
lucacarniato, line 352: i think we have defined constants in mesh2d

        static constexpr UInt m_numNodesQuads = 4;                                                 ///< Number of nodes in a quadrilateral
        static constexpr UInt m_numNodesInTriangle = 3;                                            ///< Number of nodes in a triangle

maybe we should move these constants to meshkernel::constants namespace

lucacarniato (changes requested):

libs/MeshKernel/src/Polygon.cpp
BillSenior: I made this change, but in another item

lucacarniato (approved):

PR #235, opened 2023-10-04
Feature/gridedit 502 ridge refinement

BillSenior:


libs/MeshKernel/src/SamplesHessianCalculator.cpp
```diff
@@ -0,0 +1,350 @@
...
+                                                                                            const Projection projection,
+                                                                                            UInt numberOfSmoothingIterations,
+                                                                                            const UInt numX,
+                                                                                            const UInt numY)
+{
+    std::vector<Sample> result = rawSamplePoints;
+
+    Hessian hessian(2, numY, numX);
+    PrepareSampleForHessian(rawSamplePoints, projection, numberOfSmoothingIterations, hessian);
+
```
BillSenior, line 339: Should there be a check that numX * numY == rawSamplePoints.size

BillSenior: I have added this

PR #238, opened 2023-10-06
Feature/gridedit 622 improve faces deletion

lucacarniato:


libs/MeshKernel/src/Mesh2D.cpp
```diff
@@ -1642,164 +1623,298 @@ std::tuple<meshkernel::UInt, meshkernel::UInt> Mesh2D::IsSegmentCrossingABoundar
...
+                                                                        false,
+                                                                        m_projection);
+            if (!isEdgeCrossed)
+            {
+                continue;
+            }
+
+            crossedSegmentIndex = segmentIndex;
+            crossedEdgeIndex = edgeIndex;
+            break;
```
BillSenior, line 1660: is it possible to change this to

if (isEdgeCrossed)
{
   crossedSegmentIndex = ...
   crosedEdgeIndex  = ...
   break;
}


Do you think it is easier to read?

libs/MeshKernel/src/Mesh2DIntersections.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+    UInt numSteps = 0;
+    bool intersectionFound = false;
+    double crossProductValue = constants::missing::doubleValue;
+    double adimensionalPolylineSegmentDistance = constants::missing::doubleValue;
+    double adimensionalEdgeDistance = constants::missing::doubleValue;
+    const auto polyLineSize = static_cast<UInt>(polyLine.size());
+
+    const auto checkPolyLineIndex = [&]
+    {
```
BillSenior, line 100: Does this compile ok? with no parentheses

libs/MeshKernel/src/Mesh2DIntersections.cpp
```diff
@@ -0,0 +1,363 @@
...
+    while (true)
+    {
+        // find the seed
+        const auto [isSeedFound, crossedEdgeIndex, crossedSegmentIndex] = GetIntersectionSeed(
+            m_mesh,
+            polyLine,
+            vistedEdges);
+
+        // no valid seed found in the entire mesh, we are done
+        if (!isSeedFound)
```
BillSenior, line 176: As above, may be able to remove isSeedFound. It is just a proxy for crossedEdgeIndex == nullvalue

libs/MeshKernel/src/Mesh2DIntersections.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+                    if (!intersectionFound)
+                    {
+                        const auto segmentIntersection = GetNextEdgeIntersection(polyLine, edgeIndex, segmentIndex, nextSegmentIndex, 1);
+                        intersectionFound = std::get<0>(segmentIntersection);
+                        firstIndex = std::get<1>(segmentIntersection);
+                        secondIndex = std::get<2>(segmentIntersection);
+                        crossProductValue = std::get<3>(segmentIntersection);
+                        adimensionalPolylineSegmentDistance = std::get<4>(segmentIntersection);
+                        adimensionalEdgeDistance = std::get<5>(segmentIntersection);
```
BillSenior: Could these be put in a struct? this could then be returned from tthe GetNextEdgeIntersection function and the same struct could be passed to the updateEdgeIntersections function to reduce the number of parameters

libs/MeshKernel/include/MeshKernel/Mesh2DIntersections.hpp
```diff
@@ -0,0 +1,148 @@
...
+        double edgeDistance{constants::missing::doubleValue};                        ///< The location of the intersection expressed as an adimensional distance from the edge start
+    };
+
+    /// An intersection with a mesh face
+    struct FaceMeshPolylineIntersection
+    {
+        double polylineDistance{constants::missing::doubleValue}; ///< The location of the intersection expressed as an adimensional distance from the polyline start
+        UInt faceIndex{constants::missing::uintValue};            ///< The face index
+        std::vector<UInt> edgeIndexses;                           ///< The indexes of crossed edges
+        std::vector<UInt> edgeNodes;                              ///< The indexes of the nodes defining the crossed edges
```
BillSenior, line 56: Indexes is correct, but indices is the more common usage

libs/MeshKernel/src/Mesh2DIntersections.cpp
```diff
@@ -0,0 +1,353 @@
...
+
+                    if (!intersectionFound)
+                    {
+                        std::tie(intersectionFound,
+                                 firstIndex,
+                                 secondIndex,
+                                 crossProductValue,
+                                 adimensionalPolylineSegmentDistance,
+                                 adimensionalEdgeDistance) = GetNextEdgeIntersection(polyLine, edgeIndex, segmentIndex, nextSegmentIndex, 1);
+                    }
```
BillSenior, line 230: can use a enum, e.g. direction {Forward, Backward} in place of 1 and -1

libs/MeshKernel/include/MeshKernel/Mesh2DIntersections.hpp
```diff
@@ -0,0 +1,148 @@
...
+            std::erase_if(intersections, [](const T& v)
+                          { return v.polylineDistance < 0; });
+        }
+
+    private:
+        /// @brief Gets one edge intersection
+        /// @returns The intersection seed
+        std::tuple<bool, UInt, UInt> GetIntersectionSeed(const Mesh2D& mesh,
+                                                         const std::vector<Point>& polyLine,
+                                                         const std::vector<bool>& vistedEdges) const;
```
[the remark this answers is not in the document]
lucacarniato: it is used

libs/MeshKernel/include/MeshKernel/Mesh2DIntersections.hpp
```diff
@@ -0,0 +1,148 @@
...
+                                            const UInt edgeIndex,
+                                            std::vector<FaceMeshPolylineIntersection>& intersections)
+        {
+            intersections[faceIndex].faceIndex = faceIndex;
+            intersections[faceIndex].edgeIndexses.emplace_back(edgeIndex);
+        }
+
+        Mesh2D& m_mesh;                                                      ///< The mesh where the edges should be found
+        std::vector<EdgeMeshPolylineIntersection> m_edgesIntersectionsCache; ///< A cache for saving the edge intersections of one inner or outer
+        std::vector<FaceMeshPolylineIntersection> m_facesIntersectionsCache; ///< A cache for saving the local face intersections of one inner or outer
```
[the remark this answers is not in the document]
lucacarniato: Making them local to a function would mean passing them to Compute(const std::vector<Point>& polyLine) because this function is used several times for a polygon (outer and inner rings). This make the public api more difficult to use

libs/MeshKernel/src/Mesh2DIntersections.cpp
lucacarniato: yes, it works

libs/MeshKernel/src/Mesh2DIntersections.cpp
lucacarniato: done, although it exposes the crossedEdgeIndex invalid value

libs/MeshKernel/include/MeshKernel/Mesh2DIntersections.hpp
lucacarniato: corrected

BillSenior (approved):

PR #241, opened 2023-10-12
GRIDEDIT-712 Fixed bug when connecting DD meshes

BillSenior:
Added several new unit tests.

### DIFF
--- a/libs/MeshKernel/include/MeshKernel/ConnectMeshes.hpp
+++ b/libs/MeshKernel/include/MeshKernel/ConnectMeshes.hpp
@@ -48,13 +48,18 @@ namespace meshkernel
         /// @brief The default and maximum value of the fraction of the edge length used to determine if edges are adjacent.
         static constexpr double DefaultMaximumSeparationFraction = 0.4;
 
+        /// @brief The cos of the minimum angle between two lines to be considered parallel or almost parallel.
+        ///
+        /// The angle may be close to 0 or pi.
+        static constexpr double minimumParallelDeviation = 0.9;
+
         /// @brief Connect grids.
         ///
         /// @param [in,out] mesh The mesh
         /// @param [in] separationFraction The fraction of the shortest edge to use when determining neighbour edge closeness
         /// @note separationFraction should be in the interval (0, max], where max = DefaultMaximumSeparationFraction,
         /// If the value is outside of this range then a RangeError will be thrown.
-        void Compute(Mesh2D& mesh, const double separationFraction = DefaultMaximumSeparationFraction) const;
+        static void Compute(Mesh2D& mesh, const double separationFraction = DefaultMaximumSeparationFraction);
 
     private:
         /// @brief The maximum number of hanging nodes along a single element edge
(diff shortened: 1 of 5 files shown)


libs/MeshKernel/include/MeshKernel/ConnectMeshes.hpp
ahmad-el-sayed, line 1: All methods, public and private, can be made static.

libs/MeshKernel/src/ConnectMeshes.cpp
```diff
@@ -63,6 +65,9 @@ void meshkernel::ConnectMeshes::AreEdgesAdjacent(const Mesh2D& mesh,
         {
             endNode = mesh.m_edges[edge2].second;
         }
+
+        double absCosPhi = std::abs(NormalizedInnerProductTwoSegments(edge1Start, edge1End, edge2Start, edge2End, mesh.m_projection));
+        areAdjacent = (absCosPhi > 0.9 && absCosPhi <= 1.0);
```
ahmad-el-sayed, line 70: Can the bounds be also (optional) function parameters?

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -2290,7 +2290,7 @@ namespace meshkernelapi
         return lastExitCode;
     }
 
-    MKERNEL_API int mkernel_mesh2d_connect_meshes(int meshKernelId, const Mesh2D& mesh2d)
+    MKERNEL_API int mkernel_mesh2d_connect_meshes(int meshKernelId, const Mesh2D& mesh2d, double searchFraction)
```
ahmad-el-sayed, line 2293: Change in signature of API function, must adapt MKPy, MKNet and GEP.

ahmad-el-sayed (changes requested):

libs/MeshKernel/include/MeshKernel/ConnectMeshes.hpp
BillSenior: Do you think that I should even make them free functions?

https://github.com/henrytien/effective-cpp-3rd#item-23-prefer-non-member-non-friend-functions-to-member-functions

libs/MeshKernel/src/ConnectMeshes.cpp
BillSenior: I can do this but I think that the algorithm is quite fragile and, as can be shown with the separation fraction, any small change results in garbage being generated.

But I will certainly make a named constant

libs/MeshKernel/include/MeshKernel/ConnectMeshes.hpp
BillSenior: I retract the question above. I think that this class will not be expanded, as its function is quite specific. Its algorithm may, however be improved at some point.

So, I will make all functions static

ahmad-el-sayed (approved):

libs/MeshKernelApi/src/MeshKernel.cpp
ahmad-el-sayed: MKNet and GEP do not support this functionality yet. Adapting MKPy will be done in this [PR](https://github.com/Deltares/MeshKernelPy/pull/93).

lucacarniato (approved):

PR #243, opened 2023-10-16
correct merge nodes algorithm

lucacarniato:


libs/MeshKernel/src/Mesh.cpp
```diff
@@ -276,18 +276,26 @@ void Mesh::MergeNodesInPolygon(const Polygons& polygon, double mergingDistance)
     // first filter the nodes in polygon
     std::vector<Point> filteredNodes(GetNumNodes());
```
ahmad-el-sayed, line 277: Store GetNumNodes() once as const and reuse at lines 227, 278 and 282

libs/MeshKernel/src/Mesh.cpp
```diff
@@ -395,6 +403,24 @@ void Mesh::ComputeEdgesLengths()
...
+
+    const auto isNodeInPolygon = IsLocationInPolygon(polygon, Location::Nodes);
+    for (UInt e = 0; e < numEdges; e++)
+    {
+        const auto& [firstNode, secondNode] = m_edges[e];
+        if (!isNodeInPolygon[firstNode] && !isNodeInPolygon[secondNode])
+        {
+            continue;
+        }
+        result = std::min(result, m_edgeLengths[e]);
```
ahmad-el-sayed, line 419: how about
```
if (isNodeInPolygon[firstNode] && isNodeInPolygon[secondNode])
{
    result = std::min(result, m_edgeLengths[e]);
}
```

ahmad-el-sayed (changes requested):

ahmad-el-sayed (approved):

PR #250, opened 2023-10-25
Feature/gridedit 766 improve mesh deletion speed

lucacarniato:
From about 800 seconds down to 70 seconds

### DIFF
--- a/libs/MeshKernel/include/MeshKernel/BoundingBox.hpp
+++ b/libs/MeshKernel/include/MeshKernel/BoundingBox.hpp
@@ -146,6 +146,18 @@ namespace meshkernel
         /// @brief Return the delta of the bounding box.
         Vector Delta() const;
 
+        /// @brief Create a bounding box from two points
+        template <std::derived_from<Point> T>
+        static BoundingBox CreateBoundingBox(const T& first, const T& second)
+        {
+            const auto lowerLeftX = std::min(first.x, second.x);
+            const auto lowerLeftY = std::min(first.y, second.y);
+            const auto upperRightX = std::max(first.x, second.x);
+            const auto upperRightY = std::max(first.y, second.y);
+
+            return BoundingBox({lowerLeftX, lowerLeftY}, {upperRightX, upperRightY});
+        }
+
     private:
         Point m_lowerLeft;  ///< The lower left corner of the bounding box
         Point m_upperRight; ///< The upper right corner of the bounding box
@@ -214,22 +226,10 @@ inline bool meshkernel::BoundingBox::Overlaps(const BoundingBox& other) const
 
     const auto& otherLowerleft = other.lowerLeft();
     const auto& otherUpperRight = other.upperRight();
-    if (m_upperRight.x < otherLowerleft.x)
-    {
-        return false;
-    }
-
-    if (otherUpperRight.x < m_lowerLeft.x)
-    {
-        return false;
-    }
-
-    if (m_upperRight.y < otherLowerleft.y)
-    {
-        return false;
-    }
-
-    if (otherUpperRight.y < m_lowerLeft.y)
+    if (m_upperRight.x < otherLowerleft.x ||
+        otherUpperRight.x < m_lowerLeft.x ||
+        m_upperRight.y < otherLowerleft.y ||
+        otherUpperRight.y < m_lowerLeft.y)
     {
         return false;
     }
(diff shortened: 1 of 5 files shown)


libs/MeshKernel/include/MeshKernel/BoundingBox.hpp
```diff
@@ -106,6 +106,11 @@ namespace meshkernel
                    point.y >= m_lowerLeft.y && point.y <= m_upperRight.y;
         }
 
+        /// @brief Checks if two bounding boxes overlaps
+        /// @param[in] boundingBox The input bounding box
+        /// @return True if the point if the current bounding box overlaps with another, false otherwise
```
ahmad-el-sayed, line 111: type: of the current

libs/MeshKernel/include/MeshKernel/BoundingBox.hpp
```diff
@@ -203,3 +208,31 @@ inline meshkernel::Vector meshkernel::BoundingBox::Delta() const
 {
     return Vector(m_upperRight.x - m_lowerLeft.x, m_upperRight.y - m_lowerLeft.y);
 }
+
+inline bool meshkernel::BoundingBox::Overlaps(const BoundingBox& other) const
+{
+
+    const auto& otherLowerleft = other.lowerLeft();
+    const auto& otherUpperRight = other.upperRight();
+    if (m_upperRight.x < otherLowerleft.x)
```
ahmad-el-sayed, line 217: optional: can have a single if statement  instead of 4 (use || between conditions)

libs/MeshKernel/src/Mesh2D.cpp
```diff
@@ -2019,3 +2019,39 @@ meshkernel::Mesh2D Mesh2D::Merge(const Mesh2D& mesh1, const Mesh2D& mesh2)
 
     return mergedMesh;
 }
+
+meshkernel::BoundingBox Mesh2D::GetBoundingBox() const
+{
+
+    Point lowerLeft(std::numeric_limits<double>::max(), std::numeric_limits<double>::max());
+    Point upperRight(-std::numeric_limits<double>::max(), -std::numeric_limits<double>::max());
```
ahmad-el-sayed, line 2027: There's an operator for that
```
inline meshkernel::Point meshkernel::operator-(const Point& pnt)
{
    return Point(-pnt.x, -pnt.y);
}
```
So can use: `Point upperRight = -lowerLeft`

libs/MeshKernel/src/Mesh2D.cpp
```diff
@@ -2019,3 +2019,39 @@ meshkernel::Mesh2D Mesh2D::Merge(const Mesh2D& mesh1, const Mesh2D& mesh2)
...
+
+    const auto numNodes = GetNumNodes();
```
ahmad-el-sayed, line 2029: Use in L. 2030 directly, no need to declare and assign numNodes. Even better, use range-based for loop:
```
for (auto const& node : m_nodes)
{
    lowerLeft.x = std::min(lowerLeft.x, nodes.x);
    lowerLeft.y = std::min(lowerLeft.y, nodes.y);
    upperRight.x = std::max(upperRight.x, nodes.x);
    upperRight.y = std::max(upperRight.y, nodes.y);
}
```

libs/MeshKernel/src/Mesh2DIntersections.cpp
```diff
@@ -228,9 +256,17 @@ void Mesh2DIntersections::Compute(const std::vector<Point>& polyLine)
...
     {
         cumulativeLength[i] = cumulativeLength[i - 1] + ComputeDistance(polyLine[i], polyLine[i - 1], m_mesh.m_projection);
+
+        const auto lowerLeftX = std::min(polyLine[i - 1].x, polyLine[i].x);
+        const auto lowerLeftY = std::min(polyLine[i - 1].y, polyLine[i].y);
+        const auto upperRightX = std::max(polyLine[i - 1].x, polyLine[i].x);
+        const auto upperRightY = std::max(polyLine[i - 1].y, polyLine[i].y);
+
+        polyLineBoundingBoxes[i - 1] = BoundingBox({lowerLeftX, lowerLeftY}, {upperRightX, upperRightY});
     }
```
BillSenior, line 270: I wonder is it possible to write a static function in the BoundingBox class (or perhaps Operations) that can compute if the bounding box of two lines (given by four points) overlap.  There would be more work involved in determining if the BBs overlap, but there would be less data to move around

BillSenior (approved):
Just a single question, you can ignore if you want.

ahmad-el-sayed (changes requested):

PR #259, opened 2023-11-14
GRIDEDIT-737 Moved Location enum to Definitions

BillSenior:


libs/MeshKernel/include/MeshKernel/Definitions.hpp
ahmad-el-sayed, line 1: Can you rename ToString to ProjectionToString? ToString is suitable has it been in a declared in namespace or a class.

PR #261, opened 2023-11-21
GRIDEDIT-798: Support macOS arm64

evetion:
As https://github.com/Deltares/MeshKernelPy/pull/112 fails to build on macos with `g++-11: error: unrecognized command-line option '-Xarch_arm64'`


evetion:
It's better to set `-Wno-deprecated-enum-enum-conversion`, but then I get `cc1: error: command-line option '-Wno-deprecated-enum-enum-conversion' is valid for C++/ObjC++ but not for C [-Werror]`. This in turn would require some more cmake configuration to fix, but that is outside of my expertise.

.github/workflows/build-macos.yml
```diff
@@ -59,59 +59,26 @@ jobs:
         id: paths
         run : |
           echo "build_dir=${{ github.workspace }}/build" >> $GITHUB_OUTPUT
-          echo "ext_deps_dir=${{ github.workspace }}/external_dependencies" >> $GITHUB_OUTPUT
           echo "install_dir=${{ github.workspace }}/install" >> $GITHUB_OUTPUT
 
-      # Step: Install system-provided dependencies
```
ahmad-el-sayed, line 65: Please restore these steps and disable them for macOS builds. They are needed for ubuntu builds and potentially windows (as much as I hate saying this) in the future.

.github/workflows/build-macos.yml
```diff
@@ -25,8 +25,8 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-          #- 'ubuntu-latest'
```
ahmad-el-sayed, line 28: Restore the comment, you can remove the quotation marks as well.

.github/workflows/build-macos.yml
```diff
@@ -59,59 +59,26 @@ jobs:
...
-          path: ${{ steps.paths.outputs.ext_deps_dir }}/netcdf-c/install/netcdf-c
-      
       # Step: CMake configuration
       - name: Configure
         run: >
           cmake
           -S ${{ github.workspace }}
           -B ${{ steps.paths.outputs.build_dir }}
           -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}
-          -DCMAKE_PREFIX_PATH=${{ steps.paths.outputs.ext_deps_dir }}/netcdf-c/install/netcdf-c
```
ahmad-el-sayed, line 114: Restore this and disable for macOS, it breaks the ubuntu build.

.github/workflows/build-macos.yml
```diff
@@ -138,5 +105,6 @@ jobs:
         if: always()
         with:
           name: meshkernel-${{ matrix.platform }}-${{ matrix.build_type }}
-          path: ${{ steps.paths.outputs.install_dir }}
-          if-no-files-found: error
\ No newline at end of file
+          path: |
```
ahmad-el-sayed, line 108: No need to for a pipe here. You can write `path: ${{ steps.paths.outputs.install_dir }}`.

cmake/compiler_config.cmake
```diff
@@ -2,6 +2,11 @@
 set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
+# Prevent g++-11: error: unrecognized command-line option '-Xarch_arm64'
```
ahmad-el-sayed, line 5: This needs justification in the comment

cmake/compiler_config.cmake
```diff
@@ -11,7 +16,15 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 # Add compiler-specific options and definitions per supported platform
 if (UNIX)
   if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-    add_compile_options("-Werror;-Wall;-Wextra;-pedantic;-Wno-unused-function")
+    add_compile_options("-fvisibility=hidden;-Werror;-Wall;-Wextra;-pedantic;-Wno-unused-function")
+    if(APPLE AND (CMAKE_HOST_SYSTEM_PROCESSOR MATCHES "arm64"))
+      # Be lenient on macos with arm64 toolchain to prevent Eigen -Werror=deprecated-enum-enum-conversion error
+      add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wno-deprecated-enum-enum-conversion>)
+      # Supress Boost warnings in test compile
```
ahmad-el-sayed, line 23: Typo fails spell check. The phrasing is unclear.

cmake/compiler_config.cmake
```diff
@@ -11,7 +16,15 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
...
+      add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wno-maybe-uninitialized>)
```
ahmad-el-sayed, line 24: This also needs to be justified in the comment. If the warning is not due to boost, please fix it in the meshkernel source.

ahmad-el-sayed (changes requested):

.github/workflows/build-macos.yml
```diff
@@ -59,59 +59,26 @@ jobs:
...
-        # NetCDF Dependencies m4, curl, and openssl are provided by the build machine
-        run: >
-          pwsh ${{ github.workspace }}/scripts/install_netcdf_static.ps1
-          -WorkDir ${{ steps.paths.outputs.ext_deps_dir }}/netcdf-c/work
-          -InstallDir ${{ steps.paths.outputs.ext_deps_dir }}/netcdf-c/install
-          -BuildType '${{ matrix.build_type }}'
-          -ParallelJobs 10
+      - name: Install system-provided dependencies
+        run: |
+          if [ "${{ runner.os }}" == "macOS" ]; then
```
ahmad-el-sayed, line 66: while we're at it, let's make the version coherent with the TC linux build:

- boost@1.81.0
- netcdf@4.9.1

ahmad-el-sayed (changes requested):

.github/workflows/build-macos.yml
evetion: 1.81 doesn't exist on brew it seems, 1.83 does.

evetion: But it turns out we can't set versions on brew :/

.github/workflows/build-macos-linux.yml
```diff
@@ -87,7 +83,7 @@ jobs:
 
       # Step: Build and install user-provided dependencies, executes only if no cache restored
       - name: Build and install user-provided dependencies
-        if: steps.restore-cached-external-dependencies.outputs.cache-hit != 'true'
+        if: runner.os == 'Linux' && steps.restore-cached-external-dependencies.outputs.cache-hit != 'true'
```
ahmad-el-sayed, line 86: I'd rather have `if: runner.os != 'macOS' && ...`

.github/workflows/build-macos-linux.yml
```diff
@@ -96,14 +92,14 @@ jobs:
           -BuildType '${{ matrix.build_type }}'
           -ParallelJobs 10
 
-    # Step: Cache user-provided dependencies, executes only if no cache restored
+      # Step: Cache user-provided dependencies, executes only if no cache restored
       - name: Cache user-provided dependencies
         uses: actions/cache/save@v3
-        if: steps.restore-cached-external-dependencies.outputs.cache-hit != 'true'
+        if: runner.os == 'Linux' && steps.restore-cached-external-dependencies.outputs.cache-hit != 'true'
```
ahmad-el-sayed, line 98: I'd rather have `if: runner.os != 'macOS' && ...`

.github/workflows/build-macos-linux.yml
```diff
@@ -62,22 +63,17 @@ jobs:
...
+        run: |
+          if [ "${{ runner.os }}" == "macOS" ]; then
+            brew install boost doxygen netcdf
+          elif [ "${{ runner.os }}" == "Linux" ]; then
+            sudo apt-get install libboost-all-dev doxygen
+          fi
+
       # Step: Restore cached user-provided dependencies
-      - name: Restore cached user-provided dependencies
+      - if: runner.os == 'Linux'
```
ahmad-el-sayed, line 75: I'd rather have `if: runner.os != 'macOS'`

.github/workflows/build-macos-linux.yml
```diff
@@ -1,4 +1,4 @@
-name: Build and test on macos
+name: Build and test on macos and linux
```
ahmad-el-sayed, line 1: remove and linux

ahmad-el-sayed (changes requested):

.github/workflows/build-macos.yml
ahmad-el-sayed: It seems that the minor version is not needed for boost.  Please try `brew install boost@1.81`

.github/workflows/build-macos-linux.yml
evetion:
```suggestion
        if: runner.os != 'macOS' && steps.restore-cached-external-dependencies.outputs.cache-hit != 'true'
```

ahmad-el-sayed (changes requested):

PR #272, opened 2023-12-18
GRIDEDIT-815 Added smoothness calculation for curvilinear grids

BillSenior:


libs/MeshKernel/src/CurvilinearGrid/CurvilinearGridSmoothness.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include <cmath>
+
+#include "MeshKernel/Constants.hpp"
+#include "MeshKernel/Vector.hpp"
+
+#include "MeshKernel/CurvilinearGrid/CurvilinearGridSmoothness.hpp"
+
+void meshkernel::CurvilinearGridSmoothness::Compute(const CurvilinearGrid& grid, const int direction, lin_alg::Matrix<double>& smoothness)
+{
+    smoothness.resize(grid.m_numM, grid.m_numN);
```
ahmad-el-sayed, line 37: You can use the utility function
```
    /// @brief         Resizes and fills a matrix
    /// @tparam        T        The data type of the matrix elements
    /// @tparam        storage  Matrix storage option
    /// @param[in,out] matrix   The matrix
    /// @param[in]     rows     The number of rows
    /// @param[in]     cols     The number of columns
    /// @param[in]     preserve Optional parameter that determines if the existing elements in
    ///                         the matrix should be preserved or overwritten by the provided
    ///                         or default (if not provided) fill value
    /// @param[in]     fill_value The optional fill value (defaults to the template class constructor)
    /// @return        True if the input matrix has been resized, false otherwise
    template <typename T, int storage>
    inline static bool ResizeAndFillMatrix(Matrix<T, storage>& matrix,
                                           Eigen::Index rows,
                                           Eigen::Index cols,
                                           bool preserve = false,
                                           T const& fill_value = {})
```

libs/MeshKernel/src/CurvilinearGrid/CurvilinearGridSmoothness.cpp
```diff
@@ -0,0 +1,93 @@
...
+    if (p0.IsValid() && p1.IsValid() && p2.IsValid())
+    {
+        double diffX10 = p1.x - p0.x;
+        double diffY10 = p1.y - p0.y;
+
+        double diffX21 = p2.x - p1.x;
+        double diffY21 = p2.y - p1.y;
+
+        double lengthSquared10 = diffX10 * diffX10 + diffY10 * diffY10;
+        double lengthSquared21 = diffX21 * diffX21 + diffY21 * diffY21;
```
ahmad-el-sayed, line 75: Can be all const

libs/MeshKernel/include/MeshKernel/CurvilinearGrid/CurvilinearGridSmoothness.hpp
```diff
@@ -0,0 +1,55 @@
...
+{
+
+    /// @brief Compute the smoothness of a grid.
+    class CurvilinearGridSmoothness
+    {
+    public:
+        /// @brief Compute the smoothness of the grid in the direction specified.
+        ///
+        /// @param grid Grid for which the smoothness is to be calculated.
+        /// @param direction Direction of the smoothness required.
```
ahmad-el-sayed, line 46: Use an enum rather than an int

libs/MeshKernel/src/CurvilinearGrid/CurvilinearGridSmoothness.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include "MeshKernel/Vector.hpp"
+
+#include "MeshKernel/CurvilinearGrid/CurvilinearGridSmoothness.hpp"
+
+void meshkernel::CurvilinearGridSmoothness::Compute(const CurvilinearGrid& grid, const int direction, lin_alg::Matrix<double>& smoothness)
+{
+    smoothness.resize(grid.m_numM, grid.m_numN);
+    smoothness.fill(constants::missing::doubleValue);
+
+    if (direction == 1)
```
ahmad-el-sayed, line 40: Can use a switch when enum is in place. Does it make sense to have an `Unknown` enum? in other words, is it needed? if yes, have `case default` and `case SmoothingDirection` bundled and throw an error.

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -2813,6 +2814,35 @@ namespace meshkernelapi
...
+
+            if (meshKernelState[meshKernelId].m_curvilinearGrid == nullptr)
+            {
+                throw meshkernel::MeshKernelError("The curvilinear grid id, {}, does not exist.", meshKernelId);
+            }
+
+            const meshkernel::CurvilinearGrid& grid = *meshKernelState[meshKernelId].m_curvilinearGrid;
+            lin_alg::Matrix<double> smoothnessMatrix;
+
+            meshkernel::CurvilinearGridSmoothness::Compute(grid, direction, smoothnessMatrix);
```
ahmad-el-sayed, line 2835: you'll need to cast direction from int to SmoothingDirection enum (or whatever you end up calling it). Since we will only allow 0 and 1 and since you can cast a potato if you have the will, throwing an error in `CurvilinearGridSmoothness::Compute` is a good idea. You'll prob need a a ToString enum-to-string converter for use in the exception message.

libs/MeshKernelApi/tests/src/CurvilinearGridTests.cpp
```diff
@@ -795,3 +795,51 @@ TEST_F(CartesianApiTestFixture, CurvilinearDeleteInterior_OnCurvilinearGrid_Shou
     ASSERT_NEAR(meshkernel::constants::missing::doubleValue, curvilinearGrid.node_x[72], tolerance);
     ASSERT_NEAR(meshkernel::constants::missing::doubleValue, curvilinearGrid.node_y[72], tolerance);
 }
+
+TEST(CurvilinearGrid, MakeRectangular_ComputeSmoothnessTest)
```
lucacarniato, line 799: small comment: MakeRectangular_ComputeSmoothnessTest. For the test name, i try to follow the convention: MethodName_StateUnderTest_ExpectedBehavior. See the test above this

lucacarniato (approved):

libs/MeshKernel/src/CurvilinearGrid/CurvilinearGridSmoothness.cpp
BillSenior: done

libs/MeshKernel/src/CurvilinearGrid/CurvilinearGridSmoothness.cpp
BillSenior: I think a switch here make complicates thing too much.

PR #274, opened 2024-01-08
GRIDEDIT-830 Added calculation of curvature for curvlinear grid

BillSenior:


libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -2845,6 +2846,36 @@ namespace meshkernelapi
         return lastExitCode;
     }
 
+    MKERNEL_API int mkernel_curvilinear_compute_curvature(int meshKernelId, int direction, double* curvature)
```
ahmad-el-sayed, line 2849: Please throw if curvature is null.

libs/MeshKernel/src/CurvilinearGrid/CurvilinearGridCurvature.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+        double diffX20 = p2.x - p0.x;
+        double diffY20 = p2.y - p0.y;
+
+        double diffX21 = p2.x - p1.x;
+        double diffY21 = p2.y - p1.y;
+
+        // Twice the area of the triangle formed by the points {p0, p1, p2}
+        double area2 = diffX10 * diffY21 - diffY10 * diffX21;
+        double distance01 = std::sqrt(diffX10 * diffX10 + diffY10 * diffY10);
```
ahmad-el-sayed, line 83: all these guys can be const

PR #280, opened 2024-01-12
Introduce values_type in GriddedSamples

lucacarniato:
Change GriddedSamples::values from double* to void* Add value_type field to indicate the numeric representation of the sample values


libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
```diff
@@ -33,6 +33,7 @@
 namespace meshkernel
 {
     /// @brief A class for performing bilinear interpolation on gridded samples
+    template <typename values_type>
```
ahmad-el-sayed, line 36: `template<floating_point ValuesType>` restricts the type to float, double, long double  or any extended floating-point types.

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
```diff
@@ -48,7 +49,7 @@ namespace meshkernel
                                               UInt numYCoord,
                                               const Point& origin,
                                               double cellSize,
-                                              const std::vector<double>& values);
+                                              const void* values);
```
ahmad-el-sayed, line 52: Is is possible ti use `const std::vector<ValueType>& values` instead?

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
```diff
@@ -81,21 +82,164 @@ namespace meshkernel
...
+        const Mesh2D& m_mesh;    ///< Reference to the mesh
+        UInt m_numXCoord;        ///< The number of x coordinates of the gridded data
+        UInt m_numYCoord;        ///< The number of y coordinates of the gridded data
+        Point m_origin;          ///< The coordinate of the origin
+        double m_cellSize = 0.0; ///< The grid cell size
 
         std::vector<double> m_xCoordinates; ///< The x coordinates of the grid
         std::vector<double> m_yCoordinates; ///< The y coordinates of the grid
-        std::vector<double> m_values;       ///< The gridded sample values
+        const values_type* m_values;        ///< The gridded sample values
```
ahmad-el-sayed, line 99: if you end up passing `const std::vector<ValueType>&` to the constructor, m`m_value` should be a const ref. In the original impl, a copy is made.

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
```diff
@@ -81,21 +82,164 @@ namespace meshkernel
...
         bool m_isCellSizeConstant;          ///< If the grid coordinates are specified using vectors of coordinates
     };
+
+    template <typename values_type>
+    BilinearInterpolationOnGriddedSamples<values_type>::BilinearInterpolationOnGriddedSamples(const Mesh2D& mesh,
+                                                                                              UInt numXCoord,
+                                                                                              UInt numYCoord,
+                                                                                              const Point& origin,
+                                                                                              double cellSize,
+                                                                                              const void* values) : m_mesh(mesh),
```
ahmad-el-sayed, line 109: It would be nice to break the line before the colon.

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
```diff
@@ -81,21 +82,164 @@ namespace meshkernel
...
+                                                                                                                    m_isCellSizeConstant(true)
+    {
+        m_values = static_cast<const values_type*>(values);
+    }
+
+    template <typename values_type>
+    BilinearInterpolationOnGriddedSamples<values_type>::BilinearInterpolationOnGriddedSamples(const Mesh2D& mesh,
+                                                                                              const std::vector<double>& xCoordinates,
+                                                                                              const std::vector<double>& yCoordinates,
+                                                                                              const void* values) : m_mesh(mesh),
```
ahmad-el-sayed, line 123: It would be nice to break the line before the colon.

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
```diff
@@ -81,21 +82,164 @@ namespace meshkernel
 
         /// @brief Gets the sample value at specific row and column
         /// @return The sample value
-        [[nodiscard]] double getGriddedValue(UInt columnIndex, UInt rowIndex) const
+        [[nodiscard]] values_type getGriddedValue(UInt columnIndex, UInt rowIndex) const
```
ahmad-el-sayed, line 85: Capitalize the first letter of the method's name.

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
```diff
@@ -81,21 +82,164 @@ namespace meshkernel
...
+                                                                                                                    m_numXCoord(static_cast<UInt>(xCoordinates.size())),
+                                                                                                                    m_numYCoord(static_cast<UInt>(yCoordinates.size())),
+                                                                                                                    m_xCoordinates(xCoordinates),
+                                                                                                                    m_yCoordinates(yCoordinates),
+                                                                                                                    m_isCellSizeConstant(false)
+    {
+        m_values = static_cast<const values_type*>(values);
+    }
+
+    template <typename values_type>
```
ahmad-el-sayed, line 133: General question about this method: if the number of nodes does not change, are the 3 for lops necessary?

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
```diff
@@ -81,21 +82,164 @@ namespace meshkernel
...
+    void BilinearInterpolationOnGriddedSamples<values_type>::Compute()
+    {
+        m_nodeResults.resize(m_mesh.GetNumNodes());
```
ahmad-el-sayed, line 136: Store `m_mesh.GetNumNodes()` as const ans reuse.

libs/MeshKernel/include/MeshKernel/Definitions.hpp
```diff
@@ -86,6 +86,16 @@ namespace meshkernel
         N  ///< N-direction
     };
 
+    /// @enum InterpolationValuesTypes
+    /// @brief The possible types of the values to be interpolated in the gridded sample 
+    enum class InterpolationValuesTypes
```
ahmad-el-sayed, line 91: Capitalize enums the you do not need the suffix "Type"

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
BillSenior: Should cellSize and perhaps even the members of Point class also be of ValueType?

BillSenior: Re. Ahmads comment, I think the values should be at least: `const ValueType* values`.

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
BillSenior: Or define your own concept, restricting the instantiation with which ever type you think will be required.

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
```diff
@@ -81,21 +82,164 @@ namespace meshkernel
...
+        {
+            return constants::missing::doubleValue;
+        }
+
+        const auto result = fractionalColumnIndex * fractionalRowIndex * getGriddedValue(columnIndex + 1, rowIndex + 1) +
+                            (1.0 - fractionalColumnIndex) * fractionalRowIndex * getGriddedValue(columnIndex, rowIndex + 1) +
+                            (1.0 - fractionalColumnIndex) * (1.0 - fractionalRowIndex) * getGriddedValue(columnIndex, rowIndex) +
+                            fractionalColumnIndex * (1.0 - fractionalRowIndex) * getGriddedValue(columnIndex + 1, rowIndex);
+        return result;
+    }
```
BillSenior, line 191: Question: If the class is intantiated with short- or intType, then will this compilation fail due to mixing floating-point and integer types in the same expression?

BillSenior: A possible solution could be to return double from the `getGriddedValue` function. so a static cast in this function

libs/MeshKernel/include/MeshKernel/Definitions.hpp
```diff
@@ -86,6 +86,16 @@ namespace meshkernel
...
+    {
+        shortType = 0, ///< short type
+        intType = 1,   ///< int type
+        floatType = 2, ///< float type
+        doubleType = 3 ///< double type
```
BillSenior, line 96: Just a comment.
Since the interpolation is going to be restricted to these four scalar types, could move all the implementation detail to the implementation file and also add


`template class meshkernel::BilinearInterpolationOnGriddedSamples<float>;`
`template class meshkernel::BilinearInterpolationOnGriddedSamples<double>;`
`template class meshkernel::BilinearInterpolationOnGriddedSamples<int>;`
`template class meshkernel::BilinearInterpolationOnGriddedSamples<short>;`

This may help to reduce compilation times, if this proves necessary

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
```diff
@@ -81,21 +82,164 @@ namespace meshkernel
 
         /// @brief Gets the sample value at specific row and column
         /// @return The sample value
-        [[nodiscard]] double getGriddedValue(UInt columnIndex, UInt rowIndex) const
+        [[nodiscard]] values_type getGriddedValue(UInt columnIndex, UInt rowIndex) const
         {
             const auto index = rowIndex * m_numXCoord + columnIndex;
-            return m_values[index];
+            return static_cast<values_type>(m_values[index]);
         }
```
BillSenior, line 89: Is the static case necessary here, m_values is a pointer to the values_type.

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
ahmad-el-sayed: Shouldn't the result also be of ValueType? Is it meaningful to return a double when the input is int for ex?

libs/MeshKernelApi/include/MeshKernelApi/MeshKernel.hpp
```diff
@@ -632,6 +632,26 @@ namespace meshkernelapi
         /// @returns Error code
         MKERNEL_API int mkernel_get_geometry_error(int& invalidIndex, int& type);
 
+        /// @brief Get the integer indicating the interpolation type short
+        /// @param[out] type The integer indicating the interpolation type short
+        /// @returns Error code
+        MKERNEL_API int mkernel_get_interpolation_type_short(int& type);
```
ahmad-el-sayed, line 638: We should find a way around this for all enums... Too much boiler plate code to get a value.
I don't think it would be the end of the world to maintain enums in MKPy and MKNet every time an enum is changed in the back-end. We do that already in a few places.

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
ahmad-el-sayed: sorry, missed the int and short. One way to support all integral and floating point types. 
```
template <typename T>
concept InterpolatableType= 
  std::floating_point<T> || 
  ( std::integral<T> &&
    !std::same_as<T, bool> &&
    !std::same_as<T, char> &&
    !std::same_as<T, char8_t> &&
    !std::same_as<T, char16_t> &&
    !std::same_as<T, char32_t> &&
    !std::same_as<T, wchar_t>
  );
```
Or simply if you really really know what you really really want
```
template <typename T>
concept InterpolatableType= 
  std::same_as<T, short> ||
  std::same_as<T, int> ||
  std::same_as<T, float> ||
  std::same_as<T, double>;
```
Then finally
```
template <InterpolatableType T>
void MyMethod(T value)
```

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
BillSenior: If ValueType is int, the result will be converted implicitly (if not a compilation error) to double and so should be explicitly converted to double (static_cast), since then other tokens in the expression are of type double. It seems that the gridded value is only used in expression like this, so it may simplify the code (fewer static casts)

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
BillSenior: This is a guess, @lucacarniato  will need to confirm.
One reason for this Jira item is to reduce the memory footprint, so larger data sets can be handled. If a vector is to be passed then a copy of the original data has to be made, increasing memory usage, since it will be passed to the api function as a raw pointer.

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
lucacarniato: short and float are the only requested types

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
```diff
@@ -30,10 +30,19 @@
...
+
+    template <typename T>
+    concept InterpolatableType =
+        std::same_as<T, short> ||
+        std::same_as<T, float>;
+
     /// @brief A class for performing bilinear interpolation on gridded samples
-    template <typename ValueType>
+    template <typename InterpolatableType>
     class BilinearInterpolationOnGriddedSamples : public MeshInterpolation
```
BillSenior: I do not think that Concepts work like this.

The concept definition is correct, but its use in the template is incorrect. Change to (The Scalar name can be changed to whatever you want)

template<InterpolatableType Scalar>

libs/MeshKernelApi/include/MeshKernelApi/Utils.hpp
```diff
@@ -421,10 +425,11 @@ namespace meshkernelapi
...
         }
-        return std::make_unique<meshkernel::BilinearInterpolationOnGriddedSamples<ValueType>>(mesh2d,
-                                                                                              xCoordinates,
-                                                                                              yCoordinates,
-                                                                                              griddedSamples.values);
+        return std::make_unique<meshkernel::BilinearInterpolationOnGriddedSamples<InterpolatableType>>(mesh2d,
+                                                                                                       xCoordinates,
+                                                                                                       yCoordinates,
+                                                                                                       std::span<InterpolatableType>{reinterpret_cast<InterpolatableType*>(griddedSamples.values),
+                                                                                                                                     static_cast<size_t>(griddedSamples.num_x * griddedSamples.num_y)});
```
BillSenior, line 432: I do not think that the reinterpret_cast will work. In the GriddedSamples class it is defined as being a pointer to double. You will get garbage values when accessing the values array. The values array needs to be of the correct type (int or float). Or a void* and be initialised with the correct type array.

libs/MeshKernelApi/include/MeshKernelApi/Utils.hpp
```diff
@@ -397,19 +400,20 @@ namespace meshkernelapi
...
-                                                                                                  griddedSamples.values);
+            return std::make_unique<meshkernel::BilinearInterpolationOnGriddedSamples<InterpolatableType>>(mesh2d,
+                                                                                                           griddedSamples.num_x,
+                                                                                                           griddedSamples.num_y,
+                                                                                                           origin,
+                                                                                                           griddedSamples.cell_size,
+                                                                                                           std::span<InterpolatableType>{reinterpret_cast<InterpolatableType*>(griddedSamples.values),
+                                                                                                                                         static_cast<size_t>(griddedSamples.num_x * griddedSamples.num_y)});
```
BillSenior, line 416: As we have learned recently, with null c arrays, should we check that the griddedSamples.values array is not null

libs/MeshKernelApi/include/MeshKernelApi/Utils.hpp
```diff
@@ -421,10 +425,11 @@ namespace meshkernelapi
         {
             yCoordinates[i] = griddedSamples.y_coordinates[i];
         }
-        return std::make_unique<meshkernel::BilinearInterpolationOnGriddedSamples<ValueType>>(mesh2d,
-                                                                                              xCoordinates,
-                                                                                              yCoordinates,
-                                                                                              griddedSamples.values);
+        return std::make_unique<meshkernel::BilinearInterpolationOnGriddedSamples<InterpolatableType>>(mesh2d,
```
BillSenior, line 428: Just a minor point of interest, should we also use the std::span for the x- and y-coordinate arrays. This will also save a little storage.

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
```diff
@@ -30,10 +30,19 @@
...
+#include <span>
+
 namespace meshkernel
 {
+
+    template <typename T>
+    concept InterpolatableType =
+        std::same_as<T, short> ||
+        std::same_as<T, float>;
+
```
BillSenior, line 43: Should double also be added to the list of InterpolatableTypes?
I guess, most of the time it will be double, only sometimes will it be something else

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
ahmad-el-sayed: Was too quick to push. Fixed.

libs/MeshKernelApi/include/MeshKernelApi/Utils.hpp
ahmad-el-sayed: std::span has a size method. Can use that.

libs/MeshKernelApi/include/MeshKernelApi/Utils.hpp
ahmad-el-sayed: Fixed, used concept properly

libs/MeshKernelApi/include/MeshKernelApi/Utils.hpp
ahmad-el-sayed: std::span has a size and empty methods. Can use one of those.

libs/MeshKernelApi/include/MeshKernelApi/Utils.hpp
ahmad-el-sayed: can do

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
ahmad-el-sayed: Used std::span

libs/MeshKernelApi/include/MeshKernelApi/Utils.hpp
ahmad-el-sayed: Good point. Done.

libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
ahmad-el-sayed: For reference, you can break before the colon.

BillSenior (approved):

PR #291, opened 2024-01-30
Feature/gridedit 872 undo poc 

BillSenior:


ahmad-el-sayed:
General comments:
- Use `#pragma once` instead of guards in the headers
- Class member functions should be capitalised
- Class member variables should be prefixed by `m_` rather than suffixed by `_`

PR #312, opened 2024-03-15
Add double and int type interpolation

lucacarniato:


libs/MeshKernel/include/MeshKernel/BilinearInterpolationOnGriddedSamples.hpp
```diff
@@ -38,7 +38,7 @@ namespace meshkernel
 
     /// @brief Defines the iterpolatable data types.
     template <typename T>
-    concept InterpolatableType = std::same_as<T, short> || std::same_as<T, float>;
+    concept InterpolatableType = std::same_as<T, short> || std::same_as<T, float> || std::same_as<T, double> || std::same_as<T, int>;
```
ahmad-el-sayed, line 41: break the line at or to improve readability

libs/MeshKernel/include/MeshKernel/Definitions.hpp
```diff
@@ -97,8 +97,10 @@ namespace meshkernel
     /// @brief The possible types of the values to be interpolated in the gridded sample
     enum class InterpolationValues
     {
-        shortType = 0, ///< short type
-        floatType = 1, ///< float type
+        shortType = 0,  ///< short type
+        floatType = 1,  ///< float type
+        doubleType = 2, ///< double type
+        intType = 3,    ///< double type
```
ahmad-el-sayed, line 103: doxy: int type

libs/MeshKernel/include/MeshKernel/Definitions.hpp
```diff
@@ -97,8 +97,10 @@ namespace meshkernel
     /// @brief The possible types of the values to be interpolated in the gridded sample
     enum class InterpolationValues
```
ahmad-el-sayed, line 98: I would:
- rename this to enum class InterpolationDataTypes
- capitalize the first letter and remove the suffix Type of each enumeration
- Have int type before double type (more coherent with existing enum)
```
enum class InterpolationDataTypes {
  Short = 0,
  Float = 1,
  Int = 2,  // moved up
  Double = 3
};
```

libs/MeshKernelApi/include/MeshKernelApi/MeshKernel.hpp
```diff
@@ -649,6 +649,16 @@ namespace meshkernelapi
         /// @returns Error code
         MKERNEL_API int mkernel_get_interpolation_type_float(int& type);
 
+        /// @brief Get the integer indicating the interpolation type double
```
ahmad-el-sayed, line 652: It is better to maintain enums in Python and .NET rather than bloat the API.

libs/MeshKernelApi/tests/src/Mesh2DRefinmentTests.cpp
```diff
@@ -566,7 +566,9 @@ class MeshRefinementSampleValueTypes : public ::testing::TestWithParam<meshkerne
     [[nodiscard]] static std::vector<meshkernel::InterpolationValues> GetData()
     {
         return {meshkernel::InterpolationValues::shortType,
```
ahmad-el-sayed, line 568: - Rename function to GetDataTypes
- reorder if you move int above double in the enum, even though it does not matter (you set the values of the enums)

libs/MeshKernelApi/tests/src/Mesh2DRefinmentTests.cpp
```diff
@@ -589,13 +591,16 @@ TEST_P(MeshRefinementSampleValueTypes, parameters)
 
     std::vector<short> valuesShort;
     std::vector<float> valuesFloat;
+    std::vector<double> valuesDouble;
+    std::vector<int> valuesInt;
+    const std::string filePath = TEST_FOLDER + "/data/MeshRefinementTests/gebcoIntegers.asc";
```
ahmad-el-sayed, line 596: we should really use std::filesystem here, also in ReadAscFile

libs/MeshKernelApi/include/MeshKernelApi/Utils.hpp
```diff
@@ -281,6 +281,14 @@ namespace meshkernelapi
...
             return ComputeGriddedDataSamples<float>(griddedSamples);
         }
+        if (griddedSamples.value_type == static_cast<int>(meshkernel::InterpolationValues::doubleType))
+        {
+            return ComputeGriddedDataSamples<double>(griddedSamples);
+        }
+        if (griddedSamples.value_type == static_cast<int>(meshkernel::InterpolationValues::intType))
+        {
+            return ComputeGriddedDataSamples<int>(griddedSamples);
+        }
```
ahmad-el-sayed, line 291: Reorder if you move int up in enum.  
Also see suggestion in the next comment.

libs/MeshKernelApi/include/MeshKernelApi/Utils.hpp
```diff
@@ -503,6 +511,14 @@ namespace meshkernelapi
...
             return CreateBilinearInterpolator<float>(mesh2d, griddedSamples);
         }
+        if (griddedSamples.value_type == static_cast<int>(meshkernel::InterpolationValues::doubleType))
+        {
+            return CreateBilinearInterpolator<double>(mesh2d, griddedSamples);
+        }
+        if (griddedSamples.value_type == static_cast<int>(meshkernel::InterpolationValues::intType))
+        {
+            return CreateBilinearInterpolator<int>(mesh2d, griddedSamples);
+        }
```
ahmad-el-sayed, line 521: reorder if you move int up in enum

Suggestions, you can cast once as such:

```
#include <iostream>
#include <stdexcept>

enum class InterpoaltionDataTypes {
    Short = 0,
    Float = 1,
    Int = 2,
    Double = 3
};

void PrintInterpoaltionDataType(int type) {
    switch(static_cast<InterpoaltionDataTypes>(type)) {
        case InterpoaltionDataTypes::Short :
            // do stuff
            std::cout<< "short\n";
        case InterpoaltionDataTypes::Float :
            // do stuff
            std::cout<< "float\n";
        case InterpoaltionDataTypes::Int :
            // do stuff
            std::cout<< "int\n";
        case InterpoaltionDataTypes::Double :
            // do stuff
            std::cout<< "double\n";
        default :
          throw std::runtime_error("Invalid type"); // use one of our own exceptions here 
    }
}

int main()
{
   try {
        PrintInterpoaltionDataType(0);
        PrintInterpoaltionDataType(1);
        PrintInterpoaltionDataType(2);
        PrintInterpoaltionDataType(3);
        PrintInterpoaltionDataType(4);
   } catch (std::exception const& ex) {
        std::cout << "Exception: " << ex.what() << '\n';
        return EXIT_FAILURE;
   }
   
   return EXIT_SUCCESS;
}
```

libs/MeshKernelApi/include/MeshKernelApi/MeshKernel.hpp
lucacarniato: Although having an API has its benefits. Changing the enum values as you indicated above would not result in additional pull requests in meshkernelNET and meshkernelPY.

ahmad-el-sayed (approved):

PR #320, opened 2024-03-29
GRIDEDIT-998 crash on mesh deletion

lucacarniato:


libs/MeshKernel/include/MeshKernel/BoundingBox.hpp
```diff
@@ -166,6 +166,9 @@ namespace meshkernel
     /// @brief Merge two bounding boxes into a single bounding box that will contain both of the original.
     BoundingBox Merge(const BoundingBox& b1, const BoundingBox& b2);
 
+    /// @brief Create a non overlapping bounding box
+    BoundingBox CreateEmptyBoundingBox();
+
```
BillSenior, line 171: Can this be a static function?

BillSenior: ignore above comment, it is not a member

libs/MeshKernel/src/Mesh2D.cpp
```diff
@@ -2236,20 +2236,24 @@ std::unique_ptr<Mesh2D> Mesh2D::Merge(const Mesh2D& mesh1, const Mesh2D& mesh2)
...
+    for (UInt n = 0; n < numNodes; ++n)
     {
-        lowerLeft.x = std::min(lowerLeft.x, m_nodes[e].x);
-        lowerLeft.y = std::min(lowerLeft.y, m_nodes[e].y);
-        upperRight.x = std::max(upperRight.x, m_nodes[e].x);
-        upperRight.y = std::max(upperRight.y, m_nodes[e].y);
+        if (!m_nodes[n].IsValid())
+        {
+            continue;
+        }
```
BillSenior, line 2248: Do you think we should using [[likely]] and [[unlikely]] attributes in some cases?

BillSenior (approved):
Just a small question, no action is necessary.

libs/MeshKernel/include/MeshKernel/BoundingBox.hpp
BillSenior, line 171: Can this function be made static?
Not necessary to do so.

BillSenior (approved):
Can be merged.
Only a single small question

libs/MeshKernel/include/MeshKernel/BoundingBox.hpp
BillSenior, line 171: Perhaps change name of function, it is already inside the BoundingBox class, so perhaps does not need to have BoundingBox in the name

PR #322, opened 2024-04-08
irregular to regular

lucacarniato:


libs/MeshKernel/src/Mesh2DToCurvilinear.cpp
```diff
@@ -0,0 +1,298 @@
...
+            if (newFaceIndex != missing::uintValue)
+            {
+                q.push(newFaceIndex);
+            }
+        }
+    }
+
+    const auto matrix = ComputeCurvilinearMatrix();
+    return std::make_unique<CurvilinearGrid>(matrix, m_mesh.m_projection);
+}
```
BillSenior, line 137: Could we add a move constructor to the CurvilinearGrid?
May same making a copy

libs/MeshKernel/src/Mesh2DToCurvilinear.cpp
```diff
@@ -0,0 +1,298 @@
...
+    const auto minI = *std::ranges::min_element(localI);
+    const auto minJ = *std::ranges::min_element(localJ);
+
+    Eigen::Matrix<UInt, 2, 2> matrix;
+    for (UInt i = 0; i < localI.size(); ++i)
+    {
+        if (localI[i] == minI && localJ[i] == minJ)
+        {
+            matrix(0, 0) = localNodes[i];
+        }
```
BillSenior, line 161: In the loop below is every element of the matrix assigned a value? If not then should it be initialised with some default set of values

libs/MeshKernel/include/MeshKernel/Mesh2DToCurvilinear.hpp
```diff
@@ -0,0 +1,89 @@
...
+    {
+    public:
+        /// @brief Constructor
+        /// @param[in] mesh The input unstructured mesh
+        explicit Mesh2DToCurvilinear(Mesh2D& mesh);
+
+        /// @brief Computes the curvilinear grid starting from a specific point
+        /// @param[in] point The point from which start growing the curvilinear mesh. The point must be inside a quadrangular face
+        /// @returns The curvilinear grid
+        std::unique_ptr<CurvilinearGrid> Compute(const Point& point);
```
BillSenior, line 52: Is there a way to determine this point programatically?
Is it possible to provide a function that can return such a point value

PR #328, opened 2024-04-30
Add an api for retriving polygons from faces

lucacarniato:


libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -1893,6 +1893,86 @@ namespace meshkernelapi
         return lastExitCode;
     }
 
+    MKERNEL_API int mkernel_mesh2d_get_face_polygons(int meshKernelId, int numNodes, GeometryList& facePolygons)
```
ahmad-el-sayed, line 1896: Syntactically,. C does not have references. Since this is a C API we should use a pointer. One added benefit is the ability to check if facePolygons is null and subsequently throw an exception. 
I don't think the unit tests cover such scenario.  Nothing to do for now, but something to think about.

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -1893,6 +1893,86 @@ namespace meshkernelapi
...
+        try
+        {
+            if (!meshKernelState.contains(meshKernelId))
+            {
+                throw meshkernel::MeshKernelError("The selected mesh kernel id does not exist.");
+            }
+            meshKernelState[meshKernelId].m_mesh2d->Administrate();
+            const auto numFaces = meshKernelState[meshKernelId].m_mesh2d->GetNumFaces();
+            meshkernel::UInt count = 0;
+            for (meshkernel::UInt f = 0u; f < numFaces; ++f)
```
ahmad-el-sayed, line 1908: no need for u in the assignment. better cast numFaces to UInt.

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -1893,6 +1893,86 @@ namespace meshkernelapi
...
+                    continue;
+                }
+                if (count != 0)
+                {
+                    facePolygons.coordinates_x[count] = missing::doubleValue;
+                    facePolygons.coordinates_y[count] = missing::doubleValue;
+                    count++;
+                }
+
+                for (meshkernel::UInt n = 0u; n < faceNodes.size(); ++n)
```
ahmad-el-sayed, line 1923: same

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -1893,6 +1893,86 @@ namespace meshkernelapi
...
+                {
+                    const auto currentNode = meshKernelState[meshKernelId].m_mesh2d->Node(faceNodes[n]);
```
ahmad-el-sayed, line 1925: can declare as const auto&: picks `const meshkernel::Point& meshkernel::Mesh::Node(const UInt index) const;`

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -1893,6 +1893,86 @@ namespace meshkernelapi
...
+                    facePolygons.coordinates_x[count] = currentNode.x;
+                    facePolygons.coordinates_y[count] = currentNode.y;
+                    count++;
+                }
+                const auto currentNode = meshKernelState[meshKernelId].m_mesh2d->Node(faceNodes[0]);
```
ahmad-el-sayed, line 1930: const auto&

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -1893,6 +1893,86 @@ namespace meshkernelapi
...
+    {
+        lastExitCode = meshkernel::ExitCode::Success;
+        try
+        {
+            if (!meshKernelState.contains(meshKernelId))
+            {
+                throw meshkernel::MeshKernelError("The selected mesh kernel id does not exist.");
+            }
+            meshKernelState[meshKernelId].m_mesh2d->Administrate();
+            const auto numFaces = meshKernelState[meshKernelId].m_mesh2d->GetNumFaces();
```
ahmad-el-sayed, line 1953: `static_cast<meshkernel::UInt>(meshKernelState[meshKernelId].m_mesh2d->GetNumFaces())`

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -1893,6 +1893,86 @@ namespace meshkernelapi
...
+            int numMatchingFaces = 0;
+            for (meshkernel::UInt f = 0u; f < numFaces; ++f)
```
ahmad-el-sayed, line 1955: 0u, u not needed

libs/MeshKernelApi/tests/src/ApiTest.cpp
```diff
@@ -3289,3 +3289,115 @@ TEST(Mesh2D, ConvertToCurvilinear_ShouldConvertMeshToCurvilinear)
...
+        3,
+        0,
+        1,
+        2,
+        3,
+        0,
+        1,
+        3,
+        0,
+        1};
```
ahmad-el-sayed, line 3313: I think if you add ; before }, you can move }; to a new line. Discovered that yesterday :)

libs/MeshKernelApi/tests/src/ApiTest.cpp
```diff
@@ -3289,3 +3289,115 @@ TEST(Mesh2D, ConvertToCurvilinear_ShouldConvertMeshToCurvilinear)
...
+    ASSERT_EQ(13, geometryListDimension);
+
+    meshkernelapi::GeometryList geometryList;
+    auto facesXCoordinates = std::vector<double>(geometryListDimension);
+    auto facesYCoordinates = std::vector<double>(geometryListDimension);
+    geometryList.coordinates_x = facesXCoordinates.data();
+    geometryList.coordinates_y = facesYCoordinates.data();
+    geometryList.num_coordinates = geometryListDimension;
+
+    errorCode = mkernel_mesh2d_get_face_polygons(meshKernelId, 5, geometryList);
```
ahmad-el-sayed, line 3370: To follow up on my first comment, the behaviour is undefined if `geometryListis` not allocated.

libs/MeshKernelApi/tests/src/CurvilinearGridTests.cpp
```diff
@@ -995,6 +995,9 @@ class CurvilinearLocationIndexTests : public ::testing::TestWithParam<std::tuple
         return {
```
ahmad-el-sayed, line 995: You can add `using namespace meshkernel;` at the beginning of this source and remove all the `meshkernel::` namespaces. Easier on the eye.

libs/MeshKernelApi/src/MeshKernel.cpp
lucacarniato: No need to cast to UInt, GetNumFaces return type is already UInt

libs/MeshKernelApi/src/MeshKernel.cpp
BillSenior: I think that the u (in 0u) is needed, otherwise the 0 will be interpreted as a signed integer.  Now, it could be that some/most/all compilers accept a literal 0, rather than 0u, when assigning to an unsigned int.

auto i = 0; // type int
auto j = 0u; // type unsigned int

i and j will be of different types.

BillSenior: gcc 14: accepts with no warning

unsigned int i = 0;

ahmad-el-sayed (approved):

PR #351, opened 2024-07-17
GRIDEDIT-1317 curvilinear boundary polygon

lucacarniato:


libs/MeshKernel/src/CurvilinearGrid/CurvilinearGrid.cpp
```diff
@@ -1304,9 +1310,91 @@ std::vector<CurvilinearGrid::CurvilinearFaceNodeIndices> CurvilinearGrid::Comput
...
+        for (UInt i = 0u; i < numFaceNodes; ++i)
+        {
+            const auto firstCurvilinearNodeIndex = faceIndices[i];
+            const auto nextIndex = NextCircularForwardIndex(i, numFaceNodes);
+            const auto secondCurvilinearNodeIndex = faceIndices[nextIndex];
+
+            const auto edge = std::make_pair(std::min(firstCurvilinearNodeIndex, secondCurvilinearNodeIndex),
+                                             std::max(firstCurvilinearNodeIndex, secondCurvilinearNodeIndex));
+
+            const auto it = boundaryEdges.find(edge);
```
andreasbuykx, line 1337: add a comment how this works: while iterating over edges of valid faces, boundary edges are seen once, all internal edges are seen by two faces.

libs/MeshKernel/src/CurvilinearGrid/CurvilinearGrid.cpp
```diff
@@ -1304,9 +1310,91 @@ std::vector<CurvilinearGrid::CurvilinearFaceNodeIndices> CurvilinearGrid::Comput
...
+{
+    std::vector<Point> result;
+
+    auto boundaryEdges = ComputeBoundaryEdges(lowerLeft, upperRight);
+    if (boundaryEdges.empty())
+    {
+        return result;
+    }
+
+    auto currentEdge = boundaryEdges.begin();
```
andreasbuykx, line 1361: comment: start with smallest first node

libs/MeshKernel/src/CurvilinearGrid/CurvilinearGrid.cpp
```diff
@@ -1304,9 +1310,91 @@ std::vector<CurvilinearGrid::CurvilinearFaceNodeIndices> CurvilinearGrid::Comput
...
+    auto startNode = currentEdge->first;
+    auto sharedNode = currentEdge->second;
+
+    while (!boundaryEdges.empty())
+    {
+        result.push_back(GetNode(startNode.m_n, startNode.m_m));
+        while (sharedNode != startNode)
```
andreasbuykx, line 1368: comment: iterate over connected edges, by finding edge with shared node

libs/MeshKernel/src/CurvilinearGrid/CurvilinearGrid.cpp
```diff
@@ -1304,9 +1310,91 @@ std::vector<CurvilinearGrid::CurvilinearFaceNodeIndices> CurvilinearGrid::Comput
...
+            sharedNode = currentEdge->first == sharedNode ? currentEdge->second : currentEdge->first;
+        }
+        result.push_back(GetNode(startNode.m_n, startNode.m_m));
+        boundaryEdges.erase(currentEdge);
+        if (boundaryEdges.empty())
+        {
+            return result;
+        }
+
+        result.emplace_back(constants::missing::doubleValue, constants::missing::doubleValue);
```
andreasbuykx, line 1386: add a comment: next disjunct loop starts again with smallest node index of remaining edges

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -687,6 +687,105 @@ namespace meshkernelapi
...
+                throw meshkernel::MeshKernelError("The upper right m index cannot be smaller than 0");
+            }
+
+            const auto lowerLeftNUnsigned = static_cast<meshkernel::UInt>(lowerLeftN);
+            const auto lowerLeftMUnsigned = static_cast<meshkernel::UInt>(lowerLeftM);
+            const auto upperRightNUnsigned = static_cast<meshkernel::UInt>(upperRightN);
+            const auto upperRightMUnsigned = static_cast<meshkernel::UInt>(upperRightM);
+            const auto boundaryPolygon = meshKernelState[meshKernelId].m_curvilinearGrid->ComputeBoundaryToPolygon({lowerLeftNUnsigned, lowerLeftMUnsigned},
+                                                                                                                   {upperRightNUnsigned, upperRightMUnsigned});
+            numberOfPolygonNodes = static_cast<int>(boundaryPolygon.size());
```
andreasbuykx, line 780: We do the same computation twice, first time just for getting the buffer size? That's expensive! Can't the result be cached?

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -687,6 +687,105 @@ namespace meshkernelapi
...
+            {
+                throw meshkernel::MeshKernelError("The selected mesh kernel id does not exist.");
+            }
+
+            if (!meshKernelState[meshKernelId].m_curvilinearGrid->IsValid())
+            {
+                throw meshkernel::MeshKernelError("Invalid curvilinear grid");
+            }
+
+            if (lowerLeftN < 0)
```
andreasbuykx, line 710: Those assertions will probably never fire. If you add these assertions then also add assertions that indexes cannot be larger than MaxM & MaxN, and that lower-left indexes must be less than corresponding upper-right indexes. Alternatively, min/max computations can be done here (and even indexes can be clamped).

libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -687,6 +687,105 @@ namespace meshkernelapi
...
```
andreasbuykx, line 757: see previous comment about assertions

libs/MeshKernelApi/tests/src/CurvilinearGridTests.cpp
```diff
@@ -1433,3 +1433,40 @@ TEST(CurvilinearGrid, SnapToSpline)
...
+    std::vector<double> coordinates_x(numberOfPolygonNodes);
+    std::vector<double> coordinates_y(numberOfPolygonNodes);
+
+    boundaryPolygon.coordinates_x = coordinates_x.data();
+    boundaryPolygon.coordinates_y = coordinates_y.data();
+    boundaryPolygon.num_coordinates = numberOfPolygonNodes;
+    boundaryPolygon.geometry_separator = constants::missing::doubleValue;
+
+    errorCode = mkernel_curvilinear_get_boundaries_as_polygons(meshKernelId, 0, 0, 9, 9, boundaryPolygon);
+    ASSERT_EQ(meshkernel::ExitCode::Success, errorCode);
```
andreasbuykx, line 1471: add some more assertions, e.g. of the sizes of all vectors in boundaryPolygon.

also add a test with lower-right and upper-left for example.

andreasbuykx (changes requested):
I just reviewed the code, didn't run tests.

libs/MeshKernel/src/CurvilinearGrid/CurvilinearGrid.cpp
```diff
@@ -1304,9 +1310,91 @@ std::vector<CurvilinearGrid::CurvilinearFaceNodeIndices> CurvilinearGrid::Comput
...
+    if (boundaryEdges.empty())
+    {
+        return result;
+    }
+
+    auto currentEdge = boundaryEdges.begin();
+    auto startNode = currentEdge->first;
+    auto sharedNode = currentEdge->second;
+
+    while (!boundaryEdges.empty())
```
andreasbuykx, line 1365: In case of holes, these polygons will appear in the output as extra outer boundary polygons, not as holes.

libs/MeshKernel/include/MeshKernel/CurvilinearGrid/CurvilinearGrid.hpp
```diff
@@ -308,13 +311,13 @@ namespace meshkernel
...
         {
-            const auto nodeIndex = GetNodeIndex(index);
+            const auto nodeIndex = GetGridNodeIndex(index);
             return GetNode(nodeIndex);
         }
 
         /// @brief Get the node CurvilinearGridNodeIndices from a position
         /// @return The CurvilinearGridNodeIndices
-        CurvilinearGridNodeIndices GetNodeIndex(const UInt index) const
+        CurvilinearGridNodeIndices GetGridNodeIndex(const UInt index) const
```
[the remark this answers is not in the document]
lucacarniato: UInt GetNodeIndex(const CurvilinearIndexPair& nodeIndex) const
 CurvilinearIndexPair GetIndexPair(const UInt index) const

libs/MeshKernel/src/CurvilinearGrid/CurvilinearGrid.cpp
```diff
@@ -1304,9 +1310,91 @@ std::vector<CurvilinearGrid::CurvilinearFaceNodeIndices> CurvilinearGrid::Comput
     return result;
 }
 
+std::set<CurvilinearGrid::CurvilinearEdge> CurvilinearGrid::ComputeBoundaryEdges(const CurvilinearGridNodeIndices& lowerLeft, const CurvilinearGridNodeIndices& upperRight) const
+{
+    std::vector<Point> result;
+    const auto facesIndices = ComputeFaceIndices(lowerLeft, upperRight);
+    std::set<CurvilinearEdge> boundaryEdges;
+
+    for (const auto& faceIndices : facesIndices)
```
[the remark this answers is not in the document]
lucacarniato: no longer relevant

libs/MeshKernel/src/CurvilinearGrid/CurvilinearGrid.cpp
lucacarniato: it does not have to be

libs/MeshKernelApi/src/MeshKernel.cpp
lucacarniato: it could, we do this for several other operations. I will write a follow up

libs/MeshKernel/src/CurvilinearGrid/CurvilinearGrid.cpp
lucacarniato: i will write an issue for this

lucacarniato: also, the current polygon boundary generation for unstructured mesh suffers the same problem

libs/MeshKernel/src/CurvilinearGrid/CurvilinearGrid.cpp
andreasbuykx: you're using a set, that's ordered, so you'll sort by std::pair of vertex indices which compares first, then second, I believe. If you don't care you could use std::unordered_set which may be faster to build, but starting with the smallest has the advantage that you know you are on the outer boundary of the polygon, not one of the holes.

andreasbuykx (approved):

PR #355, opened 2024-07-22
GRIDEDIT-1282 Add mkernel_polygon_count_linear_refine api for

lucacarniato:


libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -1606,6 +1606,30 @@ namespace meshkernelapi
...
+            if (!meshKernelState.contains(meshKernelId))
+            {
+                throw meshkernel::MeshKernelError("The selected mesh kernel id does not exist.");
+            }
+
+            auto const polygonVector = ConvertGeometryListToPointVector(polygonToRefine);
+
+            const meshkernel::Polygons polygon(polygonVector, meshKernelState[meshKernelId].m_mesh2d->m_projection);
+            auto const refinementResult = polygon.LinearRefinePolygon(0, firstNodeIndex, secondNodeIndex);
+
```
BillSenior, line 1623: Would it be better to use the m_projection from the MeshKernelState, rather than from the mesh2d
The algo. should then also work for curvilinear grids (if necessary)

BillSenior, line 1623: May need to do a static cast on the first- and secondNodeIndex. Conversion to meshkerne::UInt

BillSenior (approved):

PR #360, opened 2024-08-05
GRIDEDIT-1143 mesh2d is preserved after convesion to curvilinear grid

BillSenior:


libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -926,7 +926,6 @@ namespace meshkernelapi
             meshkernel::Mesh2DToCurvilinear mesh2DToCurvilinear(*meshKernelState[meshKernelId].m_mesh2d);
 
             meshKernelState[meshKernelId].m_curvilinearGrid = mesh2DToCurvilinear.Compute({xPointCoordinate, yPointCoordinate});
```
lucacarniato, line 928: We should preserve only the part that was not converted, not the entire mesh

lucacarniato (changes requested):

PR #367, opened 2024-08-30
GRIDEDIT 1410 non orthogonal face polygons

lucacarniato:


libs/MeshKernel/src/Mesh2D.cpp
```diff
@@ -1634,6 +1634,54 @@ std::vector<meshkernel::UInt> Mesh2D::GetHangingEdges() const
...
+    }
+
+    const auto numFaces = GetNumFaces();
+
+    // Pre-allocate memory for result vector and set all elements to false
+    std::vector<bool> result(numFaces, false);
+
+    // Retrieve orthogonality values
+    const std::vector<double>& metricValues = GetOrthogonality();
+
```
BillSenior, line 1658: Re: the const reference here. Is this behaviour well defined?
The function GetOrthogonality does not return a reference.

BillSenior (approved):

PR #376, opened 2024-10-17
GRIDEDIT-1452: Revert correct intersected/non-intersected faces deletion options 

lucacarniato:


BillSenior (approved):
The logic looks correct

PR #378, opened 2024-10-17
Feature/gridedit 1336 avoid double computations

BillSenior:


libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -935,30 +940,41 @@ namespace meshkernelapi
...
+                throw meshkernel::MeshKernelError("Boundary polygon array are null");
             }
 
-            auto lowerLeftNUnsigned = static_cast<meshkernel::UInt>(lowerLeftN);
-            auto lowerLeftMUnsigned = static_cast<meshkernel::UInt>(lowerLeftM);
-            auto upperRightNUnsigned = static_cast<meshkernel::UInt>(upperRightN);
-            auto upperRightMUnsigned = static_cast<meshkernel::UInt>(upperRightM);
+            if (meshKernelState[meshKernelId].m_boundariesAsPolygonCache == nullptr)
+            {
+                throw meshkernel::MeshKernelError("Polygon data has not been cached");
```
lucacarniato, line 955: throw meshkernel::MeshKernelError("Polygon data has not been cached, mkernel_curvilinear_count_boundaries_as_polygons must be called before");? but we will be facing this scenario as developers, because users will use the the wrapped meshkernel api

libs/MeshKernelApi/tests/src/ApiTest.cpp
```diff
@@ -1522,6 +1562,53 @@ TEST_F(CartesianApiTestFixture, CountNodesInPolygonMesh2D_OnMesh2D_ShouldCountAl
     ASSERT_EQ(12, numNodes);
 }
 
+TEST_F(CartesianApiTestFixture, GetNodesInPolygonMesh2D_OnMesh2D_NodeInPolygonFailures)
```
lucacarniato, line 1565: can you put some of the cache tests in a separate file, such as ApiCacheTest? ApiTest is already too large and we should split it in part

lucacarniato (changes requested):

libs/MeshKernelApi/src/CachedPointValues.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <cstring>
+#include <utility>
+
+#include "MeshKernelApi/CachedPointValues.hpp"
+
+meshkernelapi::CachedPointValues::CachedPointValues(const std::vector<meshkernel::Point>& coordinates)
+    : m_coordsX(coordinates.size()),
+      m_coordsY(coordinates.size())
+{
+    for (size_t i = 0; i < coordinates.size(); ++i)
```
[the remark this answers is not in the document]
BillSenior: I suspect the current implementation will be faster, perhaps not by much

lucacarniato (approved):

PR #382, opened 2024-11-04
GRIDEDIT-1522: Bug Fix, missing circumcenters calculations

lucacarniato:


libs/MeshKernel/src/Operations.cpp
```diff
@@ -1153,13 +1153,9 @@ namespace meshkernel
 
             auto const det = x43 * y21 - y43 * x21;
 
-            double maxValue = std::max(std::max(std::abs(x21), std::abs(y21)),
-                                       std::max(std::abs(x43), std::abs(y43)));
-            const double eps = std::max(0.00001 * maxValue, std::numeric_limits<double>::denorm_min());
-
-            if (std::abs(det) < eps)
+            if (IsEqual(det, 0.0))
```
BillSenior, line 1156: The way that IsEqual is implemented in the master, this will always return false (unless det == 0.0) So I think the previous version was the better check in this case

BillSenior (approved):

PR #384, opened 2024-11-11
GRIDEDIT-1516 Fixed bug in node classification and updated unit test

BillSenior:
Also added file missing in previous Jira item
This file was not required for normal execution of mesh kernel only help in debugging


libs/MeshKernel/src/Utilities/Utilities.cpp
```diff
@@ -0,0 +1,92 @@
...
+// All indications and logos of, and references to, "Delft3D" and "Deltares"
+// are registered trademarks of Stichting Deltares, and remain the property of
+// Stichting Deltares. All rights reserved.
+//
+//------------------------------------------------------------------------------
+
+#include "MeshKernel/Utilities/Utilities.hpp"
+#include "MeshKernel/Exceptions.hpp"
+
+void meshkernel::Print(const std::vector<Point>& nodes, const std::vector<Edge>& edges, std::ostream& out)
```
lucacarniato, line 31: Do you need to keep these functions?

libs/MeshKernel/src/Utilities/Utilities.cpp
```diff
@@ -0,0 +1,92 @@
...
+    out << "edges = zeros ( " << edges.size() << ", 2 );" << std::endl;
+
+    for (UInt i = 0; i < edges.size(); ++i)
+    {
+        out << "edges ( " << i + 1 << ", 1 ) = " << edges[i].first + 1 << ";" << std::endl;
+        out << "edges ( " << i + 1 << ", 2 ) = " << edges[i].second + 1 << ";" << std::endl;
+    }
+}
+
+void meshkernel::Print(const std::vector<double>& xNodes,
```
lucacarniato, line 58: Do you need to keep these functions?

lucacarniato (approved):

PR #387, opened 2024-11-21
Feature/gridedit 1536 add depth values

BillSenior:


libs/MeshKernelApi/include/MeshKernelApi/MeshKernel.hpp
```diff
@@ -1607,6 +1619,12 @@ namespace meshkernelapi
         /// @returns Error code
         MKERNEL_API int mkernel_mesh2d_set(int meshKernelId, const Mesh2D& mesh2d);
 
+        /// @brief Sets the bathymetry data for the mesh
+        /// @param[in] meshKernelId The id of the mesh state
+        /// @param[in] sampleData   The mesh2d bathymetry data
+        /// @returns Error code
+        MKERNEL_API int mkernel_mesh2d_set_bathymetry_data(int meshKernelId, const GeometryList& sampleData);
```
lucacarniato, line 1626: Should we consider passing an integer corresponding to an internal enumeration (e.g., enum InterpolatableType { bathy = 0, roughness = 1 }) along with a related dictionary (e.g., std::unordered_map<InterpolatableType, std::string> InterpolatableTypeNames)? By doing so, we can make the function more generic and avoid including the specific "bathymetry" type in the API name

lucacarniato (changes requested):

libs/MeshKernelApi/include/MeshKernelApi/MeshKernel.hpp
BillSenior: After discussion, setting of property values to be interpolated has been generalised

PR #389, opened 2024-11-27
GRIDEDIT-1448: use bounding boxes in curvilinear delete

lucacarniato:


BillSenior (approved):
Just a small question

PR #397, opened 2025-01-13
GRIDEDIT-673 Added copyright banner to all files where it was absent

BillSenior:


ahmad-el-sayed (approved):
This PR adds a copyright notice to c++ header and source files where it is missing. However, it does not address existing notices that need update. For example the end year in most of existing notices is often outdated. It is common to do this manually when editing a certain file.

To make the process easier it is recommended to provide a small bash or python script to automate the task.
- The script should be able to adapt all files recursively and achieve the following:
  - Add the notice where is missing, i.e. in new files where it was not added by the developer
  -  Update the year in file where the notice already exists
- Add the script to to scripts sub-directory
- The script should be run periodically either manually or scheduled with an force push in TeamCity.

PR #415, opened 2025-02-05
GRIDEDIT-1144 Node types is now an enum

BillSenior:


libs/MeshKernel/include/MeshKernel/Mesh2D.hpp
```diff
@@ -236,6 +227,12 @@ namespace meshkernel
         ///  @brief Classifies the nodes (makenetnodescoding)
         void ClassifyNodes();
 
+        /// @brief Get the node type
+        MeshNodeType GetNodeType(const UInt nodeId) const { return m_nodesTypes[nodeId]; }
+
+        /// @brief Get the node type
+        void GetNodeTypes(std::vector<MeshNodeType>& nodeTypes) const { nodeTypes = m_nodesTypes; }
```
lucacarniato, line 234: i think this is equivalent of returning a  reference (const or not depend on the use case)

const std::vector<MeshNodeType>& GetNodeTypes() const { return m_nodesTypes; }

I find this better because you do need one line less at the caller side

libs/MeshKernel/include/MeshKernel/Smoother.hpp
```diff
@@ -45,7 +48,7 @@ namespace meshkernel
         explicit Smoother(const Mesh2D& mesh);
 
         /// @brief Computes the smoother weights
-        void Compute();
+        void Compute(const std::vector<MeshNodeType>& nodeTypes);
```
lucacarniato, line 51: my initial intention was to keep compute method without arguments and pass what is needed in the constructor. it would be nice if we can have the same signature for the compute method for all algorithms

libs/MeshKernel/src/Mesh2D.cpp
```diff
@@ -676,31 +676,33 @@ void Mesh2D::InitialiseBoundaryNodeClassification()
...
+        if (intNodeType[firstNode] == -1 || intNodeType[secondNode] == -1)
         {
             continue;
         }
 
         if (m_edgesNumFaces[e] == 0)
         {
-            m_nodesTypes[firstNode] = -1;
-            m_nodesTypes[secondNode] = -1;
+            intNodeType[firstNode] = -1;
```
lucacarniato, line 686: i do not like this logic with the node type, but I do not think we can do much about it

lucacarniato (approved):

libs/MeshKernel/src/Mesh2D.cpp
BillSenior: and its good to keep the same integer values as dflowfm, making it easier to debug

BillSenior:
taken over by https://github.com/Deltares/MeshKernel/pull/420

PR #423, opened 2025-02-17
GRIDEDIT-1640 Orthogonalizing clg mesh after merging throws an exception

BillSenior:


.github/workflows/build-and-test-workflow.yml
```diff
@@ -95,7 +95,7 @@ jobs:
 
       # Step: CMake build
       - name: Build
-        run: cmake --build ${{ steps.paths.outputs.build_dir }} --config ${{ inputs.build_type }} -j
+        run: cmake --build ${{ steps.paths.outputs.build_dir }} --config ${{ inputs.build_type }} -j 4
```
lucacarniato, line 98: next time i suggest you make this part of a separate pr

libs/MeshKernel/src/Mesh2D.cpp
```diff
@@ -151,6 +157,24 @@ Mesh2D::Mesh2D(const std::vector<Point>& inputNodes, const Polygons& polygons, P
...
 }
 
+meshkernel::UInt Mesh2D::InvalidateEdgesWithNoFace()
+{
+    UInt numberInvalidated = 0;
+
+    for (UInt e = 0; e < m_edgesFaces.size(); ++e)
+    {
+
+        if (m_edgesFaces[e][0] == constants::missing::uintValue && m_edgesFaces[e][1] == constants::missing::uintValue)
```
lucacarniato, line 167: this should be 0 if both sides of the edge have no face: m_edgesNumFaces[e]

libs/MeshKernel/src/Smoother.cpp
```diff
@@ -425,6 +425,12 @@ std::tuple<double, double, double, double> Smoother::ComputeOperatorsForInterior
...
 void Smoother::ComputeOperatorsNode(UInt currentNode, const MeshNodeType nodeType)
 {
+
+    if (nodeType == MeshNodeType::Corner)
+    {
+        return;
+    }
+
     // the current topology index
     const auto currentTopology = m_nodeTopologyMapping[currentNode];
```
lucacarniato, line 436: std::ranges::fill(m_leftXFaceCenterCache, 0.0);)

libs/MeshKernel/tests/src/Mesh2DTest.cpp
```diff
@@ -510,18 +505,22 @@ TEST(Mesh2D, DeleteHangingEdge)
     std::vector<meshkernel::Point> nodes;
```
lucacarniato, line 505: since there is no hanging edge delete is this test relevant? or should be renamed to something else?

lucacarniato (commented):
+ Orthogonalize_OnInvalidMesh_ShouldThrowAMeshGeometryError now does not throw, unit tests fail
+ if i do not merge the edges for the maas grid, the orthogonalization still fails

.github/workflows/build-and-test-workflow.yml
BillSenior: This was just a quick test for initial timings, I will revert to original state.

libs/MeshKernel/tests/src/Mesh2DTest.cpp
BillSenior: The hanging edge is added a few lines after the mesh has been constructed. I have added a comment

lucacarniato (approved):

PR #446, opened 2025-04-11
GRIDEDIT-1761: Correct mkernel_mesh2d_insert_edge_from_coordinates

lucacarniato:


libs/MeshKernelApi/src/MeshKernel.cpp
```diff
@@ -2398,7 +2398,11 @@ namespace meshkernelapi
             std::vector<double> edgeLengths(meshkernel::algo::ComputeMeshEdgeLength(*meshKernelState[meshKernelId].m_mesh2d));
             constexpr auto lengthFraction = 0.01;
 
-            const auto minMeshEdgeLength = edgeLengths.empty() ? newEdgeLength : *std::ranges::min_element(edgeLengths);
+            auto filtered = edgeLengths | std::views::filter([](const double& x)
+                                                             { return x != meshkernel::constants::missing::doubleValue; });
+            auto minIt = std::ranges::min_element(filtered);
+
+            const auto minMeshEdgeLength = minIt == std::ranges::end(filtered) ? newEdgeLength : *minIt;
```
andreasbuykx, line 2405: I suggest to make a function for computing the minimum edge length, and write some unit tests for it, rather than test the insert-from-coordinates function.

libs/MeshKernelApi/tests/src/ApiTest.cpp
```diff
@@ -2169,6 +2169,54 @@ TEST_F(CartesianApiTestFixture, InsertEdgeFromCoordinates_OnNonEmptyMesh_ShouldI
...
+                                                                  4.0,
+                                                                  0.1,
+                                                                  0.0,
+                                                                  0.0,
+                                                                  10.0,
+                                                                  10.0,
+                                                                  nodeIndex);
+    ASSERT_EQ(meshkernel::ExitCode::Success, errorCode);
+    errorCode = meshkernelapi::mkernel_mesh2d_delete_node(meshKernelId, nodeIndex);
+    ASSERT_EQ(meshkernel::ExitCode::Success, errorCode);
```
andreasbuykx, line 2191: Can you add an assertion that there's at least one edge index with <0 edge length? Or make a comment about it? It is not very clear how this test relates to the fix made in the code.

andreasbuykx (commented):
These also need to be made robust against <0 edge lengths:
- MeshRefinement::ComputeEdgeBelowMinSizeAfterRefinement
- meshkernel::algo::MinEdgeLength

Will it be part of another PR?

andreasbuykx (approved):

PR #458, opened 2025-05-06
GRIDEDIT-1864 Hang on macos

BillSenior:


andreasbuykx (approved):
These changes appear quite unrelated to each other. Please document in the commit message why the std::couts are removed (left-over debugging?), and why the delta is changed in the test.

PR #474, opened 2025-07-07
GRIDEDIT-1911 Fixed bug when computing contacts from points 

BillSenior:
when the number of points is different from the number of nodes in the one d mesh.


AlexKosterDeltares (approved):
Good work!

PR #475, opened 2025-07-08
GRIDEDIT-1909 Fixed bug when deleting two adjacent blocks in CLG

BillSenior:


AlexKosterDeltares (approved):
Good idea for solving

PR #534, opened 2026-05-18
GRIDEDIT-2258 Collect interpolation nodes,

BillSenior:
Collect the nodes at which the interpolation is calculated, if the geometry list coordinates_x and _y are not null

### DIFF
--- a/libs/MeshKernel/include/MeshKernel/SampleAveragingInterpolator.hpp
+++ b/libs/MeshKernel/include/MeshKernel/SampleAveragingInterpolator.hpp
@@ -72,7 +72,8 @@ namespace meshkernel
         void Interpolate(const int propertyId, const std::span<const Point> iterpolationNodes, std::span<double> result) const override;
 
         /// @brief Interpolate the sample data set at the locationd defined.
-        void Interpolate(const int propertyId, const Mesh2D& mesh, const Location location, std::span<double> result) const override;
+        void Interpolate(const int propertyId, const Mesh2D& mesh, const Location location,
+                         std::span<double> result, std::span<double> xCoordinates, std::span<double> yCoordinates) const override;
 
         /// @brief Interpolate the sample data set at a single interpolation point.
         ///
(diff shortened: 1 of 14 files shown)


libs/MeshKernel/include/MeshKernel/SampleAveragingInterpolator.hpp
```diff
@@ -72,7 +72,8 @@ namespace meshkernel
         void Interpolate(const int propertyId, const std::span<const Point> iterpolationNodes, std::span<double> result) const override;
 
         /// @brief Interpolate the sample data set at the locationd defined.
```
RStolkerDeltares, line 74: Consider making the x/y coordinate params optional so there's no need for creating empty spans when there're no coordinates needed:

```cpp
void Interpolate(const int propertyId, const Mesh2D& mesh, const Location location,
                 std::span<double> result,
                 std::optional<std::span<double>> xCoordinates = std::nullopt,
                 std::optional<std::span<double>> yCoordinates = std::nullopt) const override;
```

libs/MeshKernel/src/SampleAveragingInterpolator.cpp
```diff
@@ -350,17 +380,28 @@ void meshkernel::SampleAveragingInterpolator::Interpolate(const int propertyId,
...
+        std::span<double> xCoordsCopy;
+        std::span<double> yCoordsCopy;
+
+        if (location == Nodes)
+        {
+            // Only save node values if Node location is selected
+            // Do a shallow copy.
+            xCoordsCopy = xCoordinates;
+            yCoordsCopy = yCoordinates;
+        }
```
RStolkerDeltares: Can be simplified when using `std::optional`:

```cpp
InterpolateAtNodes(propertyId, mesh, nodeResult,
                   location == Nodes ? xCoordinates : std::nullopt,
                   location == Nodes ? yCoordinates : std::nullopt);
```

libs/MeshKernel/src/SampleInterpolator.cpp
```diff
@@ -39,6 +39,14 @@ void meshkernel::SampleInterpolator::SetData(const int propertyId, const std::sp
     m_sampleData[propertyId].assign(sampleData.begin(), sampleData.end());
 }
 
+void meshkernel::SampleInterpolator::Interpolate(const int propertyId, const Mesh2D& mesh, const Location location, std::span<double> result) const
+{
+    // Kept empty intentionally.
+    std::span<double> xCoords;
+    std::span<double> yCoords;
+    Interpolate(propertyId, mesh, location, result, xCoords, yCoords);
```
RStolkerDeltares: Not needed when using optional coordinate params.

RStolkerDeltares (approved):

PR #543, opened 2026-08-12
Feature/gridedit 2300 add edge smoothness to get property

BillSenior:


libs/MeshKernelApi/src/MeshSmoothnessPropertyCalculator.cpp
```diff
@@ -0,0 +1,84 @@
...
+                                          geometryList.num_coordinates, Size(state, location));
+    }
+
+    std::span<double> smoothness(geometryList.values, state.m_mesh2d->GetNumEdges());
+    meshkernel::MeshSmoothness::Compute(*state.m_mesh2d, smoothness);
+
+    if (static_cast<meshkernel::UInt>(geometryList.num_coordinates) > state.m_mesh2d->GetNumEdges())
+    {
+        std::span<double> excessValues(geometryList.values, static_cast<size_t>(geometryList.num_coordinates));
+        std::ranges::fill(excessValues.subspan(state.m_mesh2d->GetNumEdges(), static_cast<meshkernel::UInt>(geometryList.num_coordinates)),
```
RStolkerDeltares: Shouldn't the subspan count be `num_coordinates - GetNumEdges()`?

BillSenior: thanks! good catch,

veenstrajelmer (commented):
I realize I mentioned to prefer self-explanatory labels that contain the location of the property. But actually, the smoothness is a property of the grid, not only of the edges, although it is computed at the edges. Same goes for orthogonality actually, this is a grid property computed on the edges, but not a property of the edges. I think the edge prefix implies that it is a property of the edges though. Let's discuss this, it might be better to call it smoothness (and orthogonality) after all.

RStolkerDeltares (approved):

PR #544, opened 2026-08-13
GRIDEDIT-2301 Global grid always extends to poles

BillSenior:


libs/MeshKernelApi/include/MeshKernelApi/MeshKernel.hpp
```diff
@@ -1556,9 +1556,8 @@ namespace meshkernelapi
         /// @brief Compute the global mesh with a given number of points along the longitude and latitude directions.
```
RStolkerDeltares: Summary should be updated as well.

RStolkerDeltares (approved):

